PR #83, opened 2026-06-16
fix(upgrader): prime wp_roles() before plugins_loaded capability query

dknauss:
## Problem

`Upgrader::upgrade_3_3_0()`'s existing-holder guard runs `get_users( array( 'capability' => 'manage_wp_sudo', ... ) )`. On **WordPress 7.0**, `WP_User_Query::prepare_query()` dereferences the raw global `$wp_roles` directly (`$wp_roles->for_site()`) for a `capability` query, with no null guard. Migrations run from `Plugin::init()` at the `plugins_loaded` hook (gated `is_admin() || WP_CLI`), where `$wp_roles` can still be null — so under **WP-CLI** (reproduced via `wp-env start` provisioning on WP 7.0 GA) the guard fatals:

```
Uncaught Error: Call to a member function for_site() on null
in wp-includes/class-wp-user-query.php:458
```

Trigger scope: stored plugin version `< 3.3.0` (routine runs) **and** the first triggering request is WP-CLI. The PHPUnit integration bootstrap initializes roles before `plugins_loaded`, so the suite was previously unaffected.

## Fix

Prime the global `$wp_roles` with a single `wp_roles()` call at the top of `maybe_upgrade()` (after the version-compare early-return, before the routine loop). `wp_roles()` lazily instantiates the global, making every present and future capability-based user query in the routines safe regardless of how early the upgrade fires. Loop-level placement (vs inside the routine) keeps the no-op path free and removes the latent trap for future routines.

The sibling `role => administrator` query does not touch `$wp_roles`; an audit found no other `plugins_loaded`-time capability queries.

Verified against `WordPress/wordpress-develop` `class-wp-user-query.php` (`prepare_query`: `global $wpdb, $wp_roles`; the capability branch calls `$wp_roles->for_site()`/`->roles` unguarded).

## Tests

- **Unit** (`tests/Unit/UpgraderTest.php`): new ordering test asserts `wp_roles()` is invoked before the `capability` `get_users()` query; default `wp_roles` stub added in `setUp()`.
- **Integration** (`tests/Integration/UpgraderTest.php`): reproduces the null-`$wp_roles` condition (try/finally restore), runs `maybe_upgrade()`, asserts no fatal + version stamped + governance caps backfilled.

### Validation on real WP 7.0 GA (wp-env)

- Full integration suite: **187 tests green**.
- With the `wp_roles();` line removed, the integration test **errors** with `for_site() on null` at `class-wp-user-query.php:458` — confirming the test catches the regression and the fix resolves it.
- Unit suite: 794 green. PHPStan: no errors. PHPCS: clean.

## Notes

PHP-only change (upgrader + tests), no browser/admin UX surface — per AGENTS.md E2E policy, full Playwright E2E is not required here. Relevant to Phase 12 (floor bump) and Phase 14 (WordPress.org readiness).

### DIFF
--- a/includes/class-upgrader.php
+++ b/includes/class-upgrader.php
@@ -75,6 +75,16 @@ class Upgrader {
 			return;
 		}
 
+		// Prime the global $wp_roles before any routine runs. Migrations run at
+		// plugins_loaded (under WP-CLI/admin), where $wp_roles may not yet be
+		// initialized. On WP 7.0+, WP_User_Query::prepare_query() dereferences
+		// the raw global directly for a `capability` query ($wp_roles->for_site()),
+		// so a null global fatals with "Call to a member function for_site() on
+		// null". wp_roles() lazily instantiates the global; calling it here makes
+		// every present and future capability-based user query in the routines
+		// below safe, regardless of how early the upgrade fires.
+		wp_roles();
+
 		// Run each applicable routine in order.
 		foreach ( self::UPGRADES as $version => $method ) {
 			if ( version_compare( $stored, $version, '<' ) && is_callable( array( $this, $method ) ) ) {
--- a/tests/Integration/UpgraderTest.php
+++ b/tests/Integration/UpgraderTest.php
@@ -185,6 +185,73 @@ class UpgraderTest extends TestCase {
 	}
 
 	/**
+	 * Regression: 3.3.0 backfill must not fatal when the global $wp_roles is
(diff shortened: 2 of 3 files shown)


dknauss:
Superseded by #84, which bundles this upgrader fix with the Phase 11 connectors work and the coupled metrics update (this fix's tests can't pass CI without that metrics update). Closing in favor of #84.

PR #102, opened 2026-06-23
fix(gate): close two action-gating completeness gaps (CVD)

dknauss:
## Summary

Investigation and fix for a coordinated-disclosure report describing **two gate-completeness gaps** in WP Sudo's action-gating model. Both findings were independently verified against source and confirmed true positives (MEDIUM — each presupposes an already-authenticated/borrowed session, not a remote bypass).

> Note: the disclosure included a suggested patch. This PR does **not** apply that external diff verbatim — the findings were re-derived from source and the fix was implemented through the project's own design-review + TDD + reviewer workflow, with a deliberately narrower scope than the report proposed.

## Affected versions (verified against tags)

These are **pre-existing** weaknesses, not introduced in 4.0.0. Verified by diffing the release tags:

- **F2 (login-session binding)** — `wp_get_session_token` is **absent in every tag from `v1.2.0` through `v4.0.0`**; no released version ever bound the sudo proof to the login session. Affects **all releases ≤ 4.0.0**.
- **F1 (effect-level gating absent on the interactive surface)** — present since the multi-surface gate model; in `v3.4.0` the gate already arms `register_function_hooks` only on `cli`/`cron`/`xmlrpc`, with admin using `intercept()` (request-pattern matching) and no effect-level backstop. Affects the multi-surface releases through **4.0.0**.
- `includes/class-sudo-session.php` is **byte-identical (whitespace-ignored) `v3.4.0` → `v4.0.0`**; `includes/class-gate.php` changed only cosmetically (`"WP Sudo"` → `"Sudo"` strings). So a prior "clean" audit of 3.4.0 reflects a sink-oriented scan (injection/XSS/nonce), not the absence of these control-level weaknesses.

Fix lands in **4.1.0** (unreleased). No DB migration; binding is enforced only when present, so pre-4.1.0 sessions are unaffected on upgrade.

## Findings

**F1 — Effect-level gating absent on the interactive surface.** The admin surface gated only by *request-pattern matching* against enumerated core pages, so a gated-equivalent destructive action invoked through a **non-enumerated handler** (e.g. a third-party `admin-post.php` route with correct cap+nonce) ran with **no sudo challenge** — even though the identical action was blocked on WP-CLI and on core's enumerated REST routes (which use *effect-level* hooks).

**F2 — Sudo proof not bound to the login session.** `verify_token()` checked only `(user, cookie-hash)`, never `wp_get_session_token()`. A captured `wp_sudo_token` cookie could be replayed from another session for the same user, and the sudo window survived logout / `WP_Session_Tokens::destroy_all()` (no `wp_logout` teardown existed).

## Fix

**F1 — interactive effect-level backstop** (`includes/class-gate.php`)
- Session-aware guard armed on `admin_init`; hard-blocks the unambiguous destructive effects (`delete_user`, `delete_plugin`, `delete_theme`, `activate_plugin`, `upgrader_pre_install`, `export_wp`) when no sudo window is active.
- **Deliberately scoped** to those effect hooks. `pre_update_option_*` filters and the `user.create`/`user.promote` hooks are **excluded** — WordPress core and legitimate plugins write those incidentally during ordinary admin loads (e.g. `validate_active_plugins()`), so guarding them would hard-block non-gated workflows. Custom REST routes and the user.create/promote paths are tracked as follow-up increments.
- Silent on allow (the enumerated flow owns `wp_sudo_action_passed`); on block fires `wp_sudo_action_blocked` on the `admin` surface with the real user ID.

**F2 — login-session binding** (`includes/class-sudo-session.php`, `includes/class-plugin.php`)
- New `_wp_sudo_session_bind` user meta = SHA-256 of the login-session token, captured at activation.
- `verify_token()` rejects a non-empty bind that no longer matches the current login session. Enforced **only when present** → existing sessions need no migration; cookie-less surfaces (CLI/cron/app-password/WPGraphQL) carry no bind and remain policy-governed.
- Session deactivated on `wp_logout`; token captured from `set_logged_in_cookie` so login-grant sessions bind correctly. `uninstall.php` cleans the new key.

## Process (per `CLAUDE.md`)
- **Pre-implementation design review** — produced the scope ruling adopted here (admin-only destructive hooks; exclude option filters; defer REST custom routes + user.create/promote; resolve audit-hook semantics).
- **TDD** — `tests/Unit/ActionGatingCompletenessTest.php` (18 tests) written failing first; existing meta-deletion counts updated for the new key.
- **Pre-commit reviewer** — APPROVED (twice: the fix, and the CI-fix follow-up).

## Verification
- `phpunit`: **831 tests, 2375 assertions — OK**
- `phpcs` (production scope): clean
- `phpstan` level 6: **No errors**
- `verify:metrics`: in sync

https://claude.ai/code/session_01NvgBLDNajHatAH9eG59LVh


dknauss:
PR #102 review summary:

**Status**
- PR: open, ready, mergeable.
- CI: all checks green.
- Findings F1/F2 look like real true positives.
- The remediation direction is sound overall.
- I would **not merge yet** without tightening one backstop edge case.

## Main blocker

### `upgrader_pre_install` is too broad but always reports `plugin.install`

The new admin backstop treats every `upgrader_pre_install` event as:

```php
$guard( 'plugin.install', __( 'Install or update plugin/theme', 'wp-sudo' ) );
```

But `upgrader_pre_install` is shared by multiple updater types, including plugins, themes, core, and language packs. That means the new backstop may:

- block non-plugin operations while reporting `plugin.install`,
- mislabel core/theme updates in audit logs,
- potentially block language-pack updates even though those are not a gated rule,
- overstate the “unambiguous destructive effects” scope.

**Recommended fix:** accept the second `$hook_extra` argument and map/allow by type/action:

```php
add_filter(
    'upgrader_pre_install',
    function ( $response, array $hook_extra = array() ) use ( $guard ) {
        if ( is_wp_error( $response ) ) {
            return $response;
        }

        $type   = $hook_extra['type'] ?? '';
        $action = $hook_extra['action'] ?? '';

        // Map only covered high-risk updater operations.
        // plugin install/update → plugin.install or plugin.update
        // theme install/update → theme.install or theme.update if rule exists
        // core update → core.update
        // translations/language packs → allow unless explicitly gated
    },
    0,
    2
);
```

Also add tests for:
- plugin install/update mapping,
- theme install/update mapping,
- core update mapping if intended,
- language/translation updater pass-through.

## Strong parts

- Binding sudo proof to `wp_get_session_token()` is the right fix for replay.
- Capturing the token from `set_logged_in_cookie` handles login auto-grant correctly.
- `wp_logout` teardown is good defense-in-depth.
- Empty-bind compatibility path is appropriate for upgrade safety.
- Backstop scope intentionally avoids noisy option/user-role hooks, which is prudent.
- Tests are meaningful and cover the key intended paths.
- The security report is unusually good: it distinguishes sink-oriented audit misses from control-completeness failures.

## Follow-up I’d require or track

1. Add an integration assertion for `WP_Session_Tokens::destroy_all()` behavior, not only logout.
2. Track custom REST effect-level backstop as a real follow-up, not just documentation.
3. Track non-enumerated `user.create` / `user.promote` separately.
4. Consider changing docs language from “no longer outlives `destroy_all()`” to “bound sessions fail once the login session is no longer valid,” unless an integration test proves the exact `destroy_all()` lifecycle.

## Verdict

**Close, but needs one targeted correction before merge:** fix the `upgrader_pre_install` classification/overblocking issue. After that, I’d be comfortable with the remediation.

dknauss:
Thanks — blocker addressed in `e76a604`.

## Blocker: `upgrader_pre_install` over-broad / always `plugin.install`

Fixed. The backstop now reads the second `$hook_extra` arg (`add_filter(..., 0, 2)`) and routes through a new `classify_upgrader_effect()` helper instead of blanket-blocking as `plugin.install`:

- `type=plugin` → `plugin.install` / `plugin.update`
- `type=theme` → `theme.install` / `theme.update`
- **everything else → pass-through (no block, no audit event)**

The language-pack case you flagged is the key one: `Language_Pack_Upgrader` sets **`language_update_type`**, not `type`, so it never matches and passes straight through — translation updates are no longer blocked or mislabelled. Core updates likewise carry no plugin/theme `type` and pass through (they're gated via the enumerated `update-core.php` flow anyway).

I verified the `hook_extra` shapes against WordPress core trunk before keying on them (per the repo's verification rules): `Plugin_Upgrader`/`Theme_Upgrader` set `type` + `action`; `Language_Pack_Upgrader` uses `language_update_type`/`language_update`. The four rule ids (`plugin.install`/`plugin.update`/`theme.install`/`theme.update`) all exist in `class-action-registry.php`.

Tests added (`test_backstop_upgrader_classifies_by_hook_extra`): asserts each plugin/theme install/update maps to its exact rule id **and** that a language-pack payload and a core/empty payload both pass through (`wp_die` fires exactly 4 times, not 6).

On scope: I chose **pass-through on non-match** rather than fail-safe-block, because this backstop is defense-in-depth behind the enumerated `update.php`/`update-core.php` flows (which already challenge normal installs), and over-blocking translations was the actual regression — fail-safe-block would reintroduce it for non-enumerated handlers with no real security gain. Happy to revisit if you'd prefer fail-closed for unknown types.

## Follow-ups

- **(4) docs wording on `destroy_all()`** — done. Softened across `CHANGELOG`/`readme.txt`/`security-model.md` to *"the window ends on logout, and a bound proof stops verifying once its login session is no longer valid (e.g. after `destroy_all()` the user is no longer authenticated)"* — dropping the overstated exact-lifecycle claim, since no integration test proves it.
- **(1) `destroy_all()` integration assertion** — tracked in the security report's prevention section as the natural next integration assertion (alongside the logout teardown). Can add it in this PR if you'd like it in-scope rather than tracked.
- **(2) custom REST effect-level backstop** and **(3) non-enumerated `user.create`/`user.promote`** — tracked as real residual gaps in `docs/security-report-2026-06-gate-completeness.md` (§ Residual / deferred gaps), with the rationale for deferring each (REST needs app-password policy mirroring; the user hooks fire on benign high-frequency paths).

## One transparency note

The same blanket `upgrader_pre_install` → `plugin.install` labeling also exists in the **non-interactive** `register_function_hooks()` path (CLI/Cron/XML-RPC), where it predates this PR. I left it out of scope here to keep the change focused on the F1/F2 findings, but it has the same mislabel/over-block characteristic (e.g. `wp language update` under a Limited CLI policy). I can align it with `classify_upgrader_effect()` in a follow-up (or this PR) if you want them consistent.

All green locally: 832 tests / 2388 assertions, PHPStan L6 clean, PHPCS clean, `verify:metrics` in sync.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #139, opened 2026-07-04
feat(widget): record session revocations in the activity dashboard

dknauss:
## Summary

- What changed? `Event_Recorder` now subscribes to `wp_sudo_session_revoked` (accepted_args 4) and stores a `session_revoked` event row: target user in `user_id` (0 for a batch revoke-all), the reason tag in the `surface` column (`users_list_row_action` / `revoke_all_ui`, rendered as `users-list` / `revoke-all`), and the operator in context as `revoked_by`. `Dashboard_Widget` gains human-readable labels for `session_revoked` ("Revoked") and the previously raw-rendered `escalation_blocked` ("Escalation"), surface-code mappings for the two reason tags, and distinct pill CSS for both event types. The four hook-enumeration tests were updated together (two were already stale — missing `escalation_blocked`) and one gained an exhaustiveness count guard so recorder-hook drift fails loudly.
- Why was this change needed? The v4.5 Phase 24 revocation UI (#131) fires `wp_sudo_session_revoked`, but the bundled recorder never subscribed — so an operator revoking sudo sessions, a prime incident-response action, left no trace in the Session Activity dashboard widget unless the optional WSAL bridge was installed. The reason tag is stored in `surface` (not context) because the widget's event query deliberately omits the context column; surface is the only slot that keeps revocation provenance visible. WP-CLI revocations do not fire the hook; that gap is now documented next to the hook signature in `docs/developer-reference.md` rather than widened.

## Validation

- [x] `composer test` — 967 tests / 2,947 assertions (4 new: two recorder payload tests incl. the zero-target batch case, two widget render tests pinning labels and asserting the raw event strings do NOT render). PHPStan L6 + Psalm and PHPCS also pass; `bin/verify-metrics.sh` reports in sync.
- [x] Relevant `npm run test:e2e*` or manual checks are described — no E2E surface changed (widget markup only gains label/CSS map entries); recorder payloads are verified against both fire sites (`includes/class-admin.php:1587`, `:2223`) by the new unit tests.

## Checklist

- [x] Linked issue or explained why none was needed — no issue; follow-up to the Phase 24 session-revocation UI (#131, hardened in #137), where the visibility gap was found during a dashboard-widget review.
- [x] Updated docs, tests, or manual testing guidance if behavior changed — CHANGELOG unreleased entry, developer-reference note (recorder behavior + CLI-gap caveat), current-metrics refresh; pre-implementation design review performed and its modifications incorporated (reason-in-surface, target-user semantics, escalation label ride-along, exhaustive enumeration tests, no filter dropdown option, document-don't-widen for CLI).
- [x] No sensitive details disclosed publicly

https://claude.ai/code/session_017TCJoA72EDAmLviMKDB5zx

---
_Generated by [Claude Code](https://claude.ai/code/session_017TCJoA72EDAmLviMKDB5zx)_


dknauss (commented):
Reviewed the focused changes for the dashboard activity widget and Event_Recorder hook coverage. The new `wp_sudo_session_revoked` subscription, payload mapping, dashboard labels/surface labels, and docs/metrics updates are consistent with the existing event-store model. I did not find blocking issues.

Local validation passed:
- `composer test` — 967 tests / 2947 assertions
- `composer lint` — pass
- `composer analyse` — pass
- `composer verify:metrics` — pass

GitHub checks are also green, including unit, integration, E2E, CodeQL, Plugin Check, docs lint, and Codecov.

Note: GitHub would not allow this authenticated account to approve because it owns the PR, so this is submitted as a review comment rather than an approving review.

PR #186, opened 2026-07-15
fix(challenge): strip provider _ajax_nonce before the full-page 2FA submit

dknauss:
Follow-up from the #185 review. The in-editor 2FA fix (`4a910a1`) excluded reserved routing/CSRF names from the serialized POST; its sibling — the **full-page** challenge's 2FA submit in `admin/js/wp-sudo-challenge.js` — had the same class of gap for `_ajax_nonce`.

## The gap
The 2FA submit builds a `FormData` from the provider-rendered form and deletes `action`/`_wpnonce` before appending WP Sudo's own — but not `_ajax_nonce`. `check_ajax_referer()` reads `$_REQUEST['_ajax_nonce']` **before** `_wpnonce`, so a provider- or `wp_sudo_render_two_factor_fields` hook-emitted `_ajax_nonce` hidden field would take precedence over WP Sudo's appended nonce and fail verification.

## Fix
Add `body.delete('_ajax_nonce')` alongside the existing deletes — parity with the in-editor modal fix and the full-page `action`/`_wpnonce` guard.

## Test
**CHAL-08** (the existing "provider hidden fields don't interfere" regression) now:
- the `wp-sudo-e2e-two-factor` fixture also renders a shadow `_ajax_nonce` hidden field, and
- the test asserts it is present and that the 2FA submit still redirects to a granted session.

RED without the delete (the bogus `_ajax_nonce` clobbers the nonce → no redirect); GREEN with it. Runs in the wp-env E2E lane (CI); JS + PHP fixture syntax verified locally.

## Playground
No blueprint changes needed: `blueprint.json` correctly pins the latest release (`v4.6.0`), and the `main`-tracking blueprints (incl. `blueprint-editor-reauth.json`) already serve current `main`.

### DIFF
--- a/admin/js/wp-sudo-challenge.js
+++ b/admin/js/wp-sudo-challenge.js
@@ -194,10 +194,15 @@
 			var body = new FormData(twofaForm);
 
 			// The Two Factor provider's authentication_page() may render hidden
-			// fields named "action" and "_wpnonce". Delete them before setting
-			// ours so the AJAX request hits our handler, not the provider's.
+			// fields named "action", "_wpnonce", or "_ajax_nonce". Delete them
+			// before setting ours so the AJAX request hits our handler, not the
+			// provider's, and carries our nonce. `_ajax_nonce` is load-bearing:
+			// check_ajax_referer() reads $_REQUEST['_ajax_nonce'] BEFORE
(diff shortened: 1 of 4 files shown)


dknauss:
@codex review

PR #188, opened 2026-07-16
fix(demo): list several valid TOTP codes in the in-editor 2FA notice

dknauss:
## Summary

The in-editor 2FA Playground demo (`blueprint-editor-2fa.json`, linked from `readme.md`) showed only the **single** TOTP code computed at editor load. Because a viewer must complete the **password** step first, that code can lapse its ±4-step (~2 min) validity window before they reach the two-factor field — so the demo often "requires a code immediately" and then rejects it, forcing an editor reload.

- **`blueprint-editor-2fa.json`** — the demo MU-plugin's `enqueue_block_editor_assets` handler now computes **five consecutive codes** (current tick + next four) and lists them in the notice:
  ```php
  $step = (int) floor( time() / 30 );
  for ( $i = 0; $i < 5; $i++ ) {
      $codes[] = (string) \Two_Factor_Totp::calc_totp( 'JBSWY3DPEHPK3PXP', $step + $i );
  }
  ```
  Notice: *"Two-factor codes (any one works right now; they roll forward so the demo stays usable for a few minutes …): A, B, C, D, E."*
- **`readme.md`** — Milestone B paragraph updated from "shows the current code" to "lists several currently-valid codes".

## Why five codes ≈ a usable window

Two Factor validates a code against `±DEFAULT_TIME_STEP_ALLOWANCE` = **±4 ticks** around submit time (`get_authcode_valid_ticktime`). Listing the current tick plus the next four keeps at least one shown code inside that window for **~4 minutes** after load — comfortably longer than the walk-through takes — after which a reload gives a fresh set.

## Verification

Verified against upstream `WordPress/two-factor` `providers/class-two-factor-totp.php`:
- `calc_totp( $key, $step_count = false, … )` — when `$step_count === false` it uses `floor( self::time() / $time_step )` (`$time_step = 30`); passing an explicit `floor(time()/30) + n` yields the current and future ticks on the same basis the validator uses (`self::time()` returns `time()`).
- `DEFAULT_TIME_STEP_ALLOWANCE = 4`.
- Codes kept as strings (`(string)` + `implode`) so leading zeros are preserved.

Decoded MU-plugin passes `php -l`; JSON valid; `composer verify:metrics` passes. The PR #187 `load-post-new.php` replay-marker teardown is untouched. No production plugin code changed.

---
_Generated by [Claude Code](https://claude.ai/code/session_017TCJoA72EDAmLviMKDB5zx)_


dknauss:
Superseded by #192. This branch predated the login-gate fix (#189) — its blueprint lacked the `get_user_metadata` short-circuit, so merging it as-is would have regressed the demo back to the login-screen 2FA dead-end, and it also conflicted with #191's readme edits. #192 re-applies the same rolling-codes idea (current step + next 4) on top of the current fixed blueprint, with the login-gate fix preserved.

PR #235, opened 2026-07-25
docs: add action_id seam to roadmap backlog (trigger-gated)

dknauss:
## What

Adds one roadmap entry (§10, "Features to consider") capturing the sole WP-Sudo-side idea worth carrying forward from the `consequential-actions` comparison: an optional per-rule `action_id` that references a **future** core Actions API identity (e.g. `user.promote` → `core/promote-user`), while keeping per-surface matching (`admin`/`ajax`/`rest`/`stash`) local to the rule.

This is the **union** the [registry-vs-Abilities decision memo](../blob/main/docs/core-actions-registry-vs-abilities-decision.md) already anticipates — WP Sudo as a *consumer* of core's identity layer while remaining the enforcement + stash-and-replay layer core defers.

The entry explicitly records what the seam is **not**:
- Not wholesale adoption of CA's pure-data registry — that would discard the request-pattern matching that powers stash-and-replay (a strict regression; the CA demo cannot replay).
- Not adoption of CA's semantic metadata (`consequence_class`/`scope`/`annotations`) — dead data no enforcement path here reads.

Marked **YAGNI until triggered**: build only if/when core actually ships an Actions API to key against.

## Scope

Docs-only (`docs/ROADMAP.md`), one entry. No code, no behavior change.

## ⚠️ Merge note

This branch was cut from `dce81b4` and anchors the entry in the **old** §10 structure. A concurrent restructure of `ROADMAP.md` to a forward-only Now/Next/Later format is in flight on `docs/roadmap-restructure`. Expect a conflict in §10 — on merge, this entry should be **re-homed** into the new "Later" section (or `sudo-design-notes.md`) rather than merged as-is. Sequencing after the restructure lands is cleanest.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,7 +95,7 @@ composer sbom                 # Regenerate CycloneDX SBOM (.sbom/bom.json)
 ./vendor/bin/phpunit --filter testMethodName           # Run a single test method
 ```
 
-No build step. No production dependencies — only dev dependencies (PHPUnit 9.6, Brain\Monkey, Mockery, VIP WPCS, PHPStan, CycloneDX). `config.platform.php` is set to `8.1.99` so the lock file resolves packages compatible with PHP 8.1+ regardless of local PHP version.
+No build step. No production dependencies — only dev dependencies (PHPUnit 9.6, Brain\Monkey, Mockery, VIP WPCS, PHPStan, CycloneDX). `config.platform.php` is set to `8.2.99` so the lock file resolves packages compatible with PHP 8.2+ regardless of local PHP version.
 
 ## WP-CLI and remote-site safety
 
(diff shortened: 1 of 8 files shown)


dknauss:
Superseded by #236 (`docs/roadmap-restructure`, Pass 1 + Pass 2 of the ROADMAP restructure).

This PR's only unique content over its `dce81b4` base — the `action_id` seam entry — is already re-homed into the new structure at `docs/sudo-design-notes.md` (the §10 "Features to consider" analysis moved there when §10 was promoted to a standalone doc). Everything else (the Now/Next/Later restructure and the AGENTS/CLAUDE `8.1→8.2` fix) originates from that same `dce81b4` commit and is included in the restructure branch. Per this PR's own merge note — "this entry should be re-homed into the new structure rather than merged as-is" — closing in favor of the restructure PR. No content lost.

PR #268, opened 2026-07-25
feat(governance): scoped break-glass recovery + Site Health critical status (#240)

### LINKED ISSUE #240: Break-glass recovery follow-ups: scoped single-user constant + Site Health critical status
Support define( WP_SUDO_RECOVERY_MODE, <user_id_or_login> ) so break-glass grants only that single user (today wp_sudo_is_recovery_mode() treats the constant as boolean-only), and add a pull-based Site_Health::test_recovery_mode() critical status registered beside test_stale_sessions.

Source: docs/ROADMAP.md §12.1 Phase R2 (recovery-mode items 2 and 5).

_Filed from the WP Sudo roadmap backlog (Pass 3: roadmap → issues)._

dknauss:
## What & why

Closes #240. Break-glass recovery (`WP_SUDO_RECOVERY_MODE`) was boolean-only: while defined, **every** `manage_options` holder regained full Sudo governance. That is a wider blast radius than break-glass needs — the contained failure is "one admin locked out," not "all admins lost governance." This PR adds a **scoped** form and a **Site Health** signal so a forgotten constant is caught on the health dashboard, not only on a Sudo screen.

Pre-implementation design review ran first (per CLAUDE.md) and caught a **fail-open trap** — a `?int` resolver that would have read an unresolvable/typo'd target as "unscoped → grant all admins." The final design derives the grant decision from an explicit three-state classification instead.

## Changes

**Scoped recovery — `includes/functions-governance.php`**
- `WP_SUDO_RECOVERY_MODE` now accepts a **user ID or login**; boolean `true` stays the legacy unscoped form.
- `wp_sudo_recovery_mode_is_unscoped()` — strict `=== true`, so `define(..., 1)` is user ID 1, not "on" (avoids the `1 == true` loose-comparison footgun).
- `wp_sudo_recovery_mode_user(): ?int` — resolves int→ID, string→login via `get_user_by`; `null` when undefined/unscoped/empty/unresolvable. Does only a user lookup (no cap checks), so it cannot re-enter `map_meta_cap`.
- `wp_sudo_user_matches_recovery(int): bool` — the single grant seam: inactive→nobody; unscoped→current-user-only (preserves the pre-scoping contract); scoped→resolved target only; **unresolvable target matches NOBODY** (fail-closed, never falls back to any-admin).
- `wp_sudo_is_recovery_mode()` stays a loose "active at all" test. Both `wp_sudo_can()` and `wp_sudo_map_governance_meta_cap()` route through the seam; the role gate (`manage_options` / `manage_network_options`) still applies on top.

**Site Health — `includes/class-site-health.php`**
- New `test_recovery_mode()` (`direct`, pull-based) → **critical** whenever recovery is active, naming the scope (unscoped / resolved target / "resolves to no user — typo"). Target is `esc_html`'d and status-only, never in `debug_information`.

**Notice — `includes/class-admin.php`**: copy branches scoped vs unscoped vs unresolvable.

**Docs**: security-model, FAQ, developer-reference, CHANGELOG. `psalm-baseline.xml` regenerated for the new global-fn-from-namespace calls (existing project convention).

## Tests

- **Unit (1148 pass):** three-state grant logic incl. the fail-closed unresolvable case; re-pointed existing recovery tests to the new seam; SiteHealth register-count (4→5 / 5→6) and Admin notice branches.
- **Integration (run in CI — local DB unavailable):** real login/ID resolution, target-grant vs non-target-deny, and the fail-closed unresolvable path.
- Gates green: `composer test:unit`, `composer lint`, `composer analyse` (PHPStan L6 + Psalm).

## Notes for review

- `@since 5.0.0` matches this issue's milestone; `WP_SUDO_VERSION` is still `4.8.0` — version-sync happens at release per CLAUDE.md. If the next tag is `4.9.0`, the `@since` tags + developer-reference "since 5.0.0" lines need a sweep.
- Documented edge: string `'1'` resolves as **login** `1`, not user ID 1 (integer→ID, string→login); fails closed if no such login.


dknauss:
Codex review triaged and addressed in `1fb036d` (one verify-and-triage pass): conditional wording for ineligible scoped targets (Site Health + notice), readme break-glass section updated for the scoped form, removed the nonexistent `wp cli site-health` reference, added the three recovery helpers to VERSIONING.md, regenerated the POT (verify:i18n clean), and cited the live `get_user_by` source in the commit message. Resolving the threads.

Issue #262, opened 2026-07-25
Complete the in-editor sudo session-status indicator (client UI)


dknauss:
## Summary
The in-editor sudo session-status indicator is **half-shipped**: the server-side data feed landed in #204 (`bd0aab6`), but the **client-side UI was never built**, and its completion fell out of the roadmap during the forward-only restructure (#258). This issue tracks finishing it.

## What shipped (#204)
- Server-side data feed: reuses `Sudo_Session::time_remaining`, localizes `remaining` at page load (gated on `is_active()`) and returns it from the three authenticated grant responses.
- Design brief: [`.planning/in-editor-session-indicator-design-brief.md`](.planning/in-editor-session-indicator-design-brief.md).

## What remains (client UI, per the design brief)
- [ ] **Snackbar baseline** — announce-once `core/notices` snackbar (WP 6.4+ floor).
- [ ] **Persistent indicator** — feature-detected `PluginSidebar` / header countdown indicator (unified editor extensibility API, **WP 6.6+**; feature-detected so the 6.4 floor degrades to the snackbar).
- [ ] New `admin/js/` module (none exists today — `wp-sudo-editor-reauth.js` is the reauth *modal*, not the indicator).
- [ ] Tests + a `docs/current-metrics.md` update if surface/asset counts change.

## Notes
- Grace window intentionally reads inactive; session meta is network-global (per the #204 brief).
- The ROADMAP's "Gutenberg Block Editor Integration — Shipped" refers to the reauth **modal** (Milestones A/B, v4.6/4.7), **not** this indicator.

Refs: #204, #182 (research todo).

PR #38, opened 2026-05-11
[codex] Fix front-end admin bar sudo deactivation

dknauss:
## Summary

Fixes the admin bar deactivation flow when the Sudo toolbar item is clicked from the front-end.

## Root Cause

The toolbar item could render on front-end pages, but the deactivation handler was registered on `admin_init`. Clicking the toolbar item from the front-end navigated with the deactivation query string, but WordPress never reached the handler, so the sudo session stayed active.

## Changes

- Register the deactivation handler on `init` so it can process front-end and admin requests.
- Route the toolbar deactivation URL through an admin endpoint and carry the current page as `wp_sudo_redirect_to`.
- Reconstruct current URLs from request host/scheme/URI and sanitize them as full URLs to preserve encoded paths and queries.
- Add unit coverage for the new hook, redirect parameter, href shape, and cleaned redirect target.
- Add ABAR-03 E2E coverage for front-end toolbar deactivation.

## Validation

Local runtime validation available here:

- `git diff --check` passed.
- `git diff --cached --check` passed.
- `node_modules/.bin/playwright test --config tests/e2e/playwright.config.ts --project=chromium tests/e2e/specs/admin-bar-deactivate.spec.ts --list` passed and listed ABAR-01, ABAR-02, and ABAR-03.

Blocked locally by runtime availability:

- PHPUnit/PHPCS/PHPStan cannot run in this Codex runtime because `php` is not installed.
- Browser E2E cannot run in this Codex runtime because `docker` is not installed and `wp-env start` fails with `spawn docker ENOENT`.

GitHub Actions should run the PHP and Docker-backed validation for this PR.

PR #298, opened 2026-07-26
feat(editor): show sudo state on the in-editor header padlock (#288)

### LINKED ISSUE #284: Add in-editor session-status indicator screenshots to readme (#262 follow-up)
The in-editor sudo session-status indicator (#262) shipped in #277 (`fee342f`), but the README/readme **screenshot refresh was deferred as a follow-up** (per AGENTS.md UI screenshot policy) and never filed — so no screenshot currently shows the new surface. The Codex "screenshot refresh" thread on #277 was resolved by *logging* this task, not by adding an image.

Current state: `readme.md` / `readme.txt` screenshots run `.wordpress-org/screenshot-1.png` … `screenshot-10.png`. The nearest are screenshot-8 (admin-bar countdown) and screenshot-10 (in-editor 2FA **modal**) — **neither shows the indicator**.

## What to capture (needs a browser/dev-provisioned session)
The indicator only renders in a live editor with an active sudo session, so this requires a browser-capable session (e.g. `claude-playwright`) against a running WP 6.6+ editor:

1. **PluginSidebar countdown (Part B, WP 6.6+):** the "Sudo" editor sidebar panel showing the live **"Sudo active — M:SS remaining"** text (`.wp-sudo-indicator-panel`, registered as `registerPlugin('wp-sudo-session-indicator')`).
2. **Announce-once snackbar (Part A):** the grant snackbar **"Reauthenticated — sudo active for about N min."** (`core/notices`, id `wp-sudo-session-active`).

## Deliverables
- Add `.wordpress-org/screenshot-11.png` (sidebar countdown) and, if shown separately, `screenshot-12.png` (snackbar).
- Reference them with descriptive alt text in the `readme.md` Screenshots table **and** the `readme.txt` `== Screenshots ==` section.
- Bump the screenshot/asset count in `docs/current-metrics.md` if it enumerates screenshots.

## Source refs (verified)
- `admin/js/wp-sudo-session-indicator.js` — snackbar text + `SNACKBAR_ID`, `.wp-sudo-indicator-panel`, `registerPlugin('wp-sudo-session-indicator')`.
- `readme.md` Screenshots table currently ends at `screenshot-10.png`.

_Follow-up to #262 / #277._

### LINKED ISSUE #288: In-editor header indicator: add an at-a-glance active-state visual (not just the accessible-name flip)
## Problem

The in-editor sudo session indicator shipped in #262 / #277 ([`admin/js/wp-sudo-session-indicator.js`](admin/js/wp-sudo-session-indicator.js)) registers a `PluginSidebar` named "Sudo" with a static **`shield`** dashicon in the block/site-editor header. On the active↔inactive transition it flips only the button's **accessible name** (`title`: "Sudo · active" / "Sudo · inactive", lines ~238–246) — deliberately, to avoid a per-second live region.

Two problems:
1. **No at-a-glance visual cue.** A sighted user sees the same static shield whether or not sudo is active; the live `M:SS` only appears *inside* the opened panel (lines ~257–266). Below WP 6.6 there is no persistent element at all — only the transient `core/notices` snackbar (Part A).
2. **Inconsistent iconography.** The shield is an arbitrary glyph unrelated to the rest of the plugin. Everywhere else the *persistent* indicator is the admin-bar chip, which uses a **`dashicons-unlock`** padlock — and an unlocked padlock is the more meaningful symbol (unlocked = elevated/sudo-active).

## Proposed follow-up: reuse the admin-bar padlock, identically

Drop the shield. Make the editor header control the **same green/red unlocked-padlock chip the admin bar uses**, so the two surfaces read as one system.

Admin-bar reference (`admin/css/wp-sudo-admin-bar.css`, `includes/class-admin-bar.php`, `admin/js/wp-sudo-admin-bar.js`):
- Glyph: `dashicons-unlock`.
- **Active:** background `#2e7d32` (hover `#1b5e20`), white glyph/text.
- **Expiring (final 60 s):** `.wp-sudo-expiring` → background `#c62828` (hover `#b71c1c`), white.
- The red transition triggers at `remaining <= 60`.

Editor changes:
- Swap `icon: 'shield'` → the `dashicons-unlock` glyph.
- Give the pinned header button the green (`#2e7d32`) active background and the red (`#c62828`) final-minute background, matching the admin-bar tokens exactly (white glyph on the chip). Inactive/expired reverts to the default (no chip).
- Drive the red transition off the module's existing absolute-deadline state — mirror the admin bar's `remaining <= 60` → expiring class. One state change, not per-second, so it stays a11y-quiet (keep the accessible-name flip as the AT signal; the grant snackbar remains the single spoken announcement — do **not** add the admin bar's 60/30/10 s SR milestones here).
- Optionally surface a compact `M:SS` (or "~N min") on/near the header control so time-remaining is visible without opening the panel (re-derived from the existing deadline; no new endpoint, no polling).
- Keep the WP 6.4–6.5 snackbar fallback unchanged (no PluginSidebar there).
- Respect `prefers-reduced-motion` (mirror the admin bar, which skips its pulse animation).

**Implementation caveat to verify:** confirm Gutenberg re-renders the *pinned header* button when the `PluginSidebar` icon/class changes on each tick (the panel component already re-renders). If it does not reflect a live icon-prop change, fall back to toggling a CSS class on the pinned button (targetable by its stable `aria-label`/name) via a small effect.

## Acceptance
- Editor header control uses the `dashicons-unlock` padlock and is visually **identical in color/state to the admin-bar chip** (green active, red final-minute), persisting while the panel is closed, on WP 6.6+ (post + site editor).
- Red state appears in the final 60 s and reverts correctly on re-grant/expiry.
- No per-second live-region announcements; the grant snackbar remains the single spoken announcement; accessible-name flip retained.
- Capture the missing #262 screenshots (closes the #284 gap) once the visual lands.

## References
- Module: [`admin/js/wp-sudo-session-indicator.js`](admin/js/wp-sudo-session-indicator.js)
- Admin-bar parity source: `admin/css/wp-sudo-admin-bar.css`, `includes/class-admin-bar.php`, `admin/js/wp-sudo-admin-bar.js`
- Design brief: `.planning/in-editor-session-indicator-design-brief.md`
- Prior work: #262, #277 · deferred screenshot follow-up: #284

dknauss:
## Summary

**What changed?** The in-editor indicator's pinned header button used a static `shield` dashicon whose only state signal was its accessible name. It now wears the admin-bar chip's own unlocked padlock and tokens — green `#2e7d32` while active, red `#c62828` in the final 60 s (the admin bar's own `remaining <= 60` threshold), reverting to the stock Gutenberg button at expiry.

**Why?** A sighted user editing full-screen — where the admin bar and its countdown never render — saw an identical glyph whether sudo was active, about to expire, or inactive. Not a security hole; a state-legibility failure on the one surface that hides the admin bar. The shield was also iconography used nowhere else in the plugin, so the two surfaces did not read as one system.

State rides a class on `<body>` toggled by a module-level subscriber guarded on a derived three-value state string, so it fires on transitions only, never per second.

<details>
<summary>Why a body class, and why module-level</summary>

- **Body class, not a class on the button** — Gutenberg re-renders that button with a fresh `className` whenever `is-pressed` flips on panel open/close, wiping anything added from outside React.
- **Body class, not `:has()`** — `:has()` post-dates the plugin's WP 6.4 floor in Firefox (121, Dec 2023).
- **Module-level, not a `useEffect`** — the panel re-renders every second and the site editor is an SPA where `document.body` outlives route changes, so a component-scoped effect has an unmount path that must be exactly right or a stale chip survives. A `lastState` guard has no unmount to get wrong.
</details>

### Design-review resolutions

The brief's five open questions were settled by a design reviewer before implementation and are recorded in `.planning/288-editor-padlock-indicator-design-brief.md`. Three decisions changed the issue's spec:

- **A third accessible name, "Sudo · expiring."** One extra transition, and a name change on an unfocused button is not announced, so it costs no interruption.
- **The expiring state also swaps the glyph** (to `warning`). `#2e7d32` and `#c62828` are **1.09:1 against each other**, so on an icon-only button green-vs-red is a WCAG 1.4.1 colour-only distinction — the same swatch under a red-green deficiency, in greyscale, and in forced-colors mode. An accessible name cannot discharge 1.4.1, a name being programmatic rather than visual. The admin bar needs no equivalent because it carries visible `Sudo: M:SS` text.
- **No `transition`, unlike the admin bar.** The green→red cross-fade was observed parked at `currentTime` 0 while the page was not producing frames, leaving the chip showing a **stale colour**. An expiry cue that can lie is worse than no animation, and this is a discrete security state, not a decorative one. With no motion there is also nothing for a `prefers-reduced-motion` block to reduce.

Also: no `!important` (the selectors are (0,4,2); core's competing rules top out at (0,3,0), verified in the live cascade); `:not(.is-pressed)` so Gutenberg keeps its panel-open affordance; the optional compact `M:SS` omitted, because on the pinned toggle the only slot for text is `title`, which *is* the accessible name.

Contrast against the white glyph: **5.13:1** active, **7.87:1** active-hover, **5.62:1** expiring, **6.57:1** expiring-hover — all clear WCAG 1.4.11 (3:1, the criterion governing a meaningful icon) and the stricter 1.4.3 (4.5:1).

Below WP 6.6 nothing changes: no pinned button exists, the module falls back to the Part A snackbar, and the stylesheet's selectors match nothing.

### Also closes #284

Adds `.wordpress-org/screenshot-11.png` (indicator panel with live countdown) and `screenshot-12.png` (green padlock chip + grant snackbar), with `readme.md` / `readme.txt` entries and a reproducible capture step in `npm run screenshots`. The capture seeds via the grant event rather than acquiring a real session — with a live session this editor screen stops producing frames under headless Chromium and every capture path times out.

## Validation

- [x] `composer test` — 1151 tests, 3396 assertions
- [x] `composer lint`, `composer analyse`, `composer verify:metrics` — all clean
- [x] E2E against a live WP 7.0 wp-env: 8/8 in `editor-session-indicator.spec.ts`, both new visual baselines (VISN-05/06)

New coverage: **INDICATOR-06** (glyph is `dashicons-unlock`, not `shield`), **INDICATOR-07** (green while active, red at `remaining <= 60`, reverts to stock at expiry leaving no body class, re-paints on a later grant), **INDICATOR-08** (expiring is distinguishable without colour), **VISN-05/06** element snapshots, and two `PluginTest` cases for the stylesheet enqueue (present when logged in, absent for anonymous).

Both new E2E specs live in files already wired into the `e2e.yml` matrix — no orphaned spec file.

> [!NOTE]
> The four pre-existing visual baselines (VISN-01…04) fail locally on macOS both with **and without** this change — verified on a stashed tree. That is the documented macOS-vs-Linux snapshot drift, in a `continue-on-error` workflow, and is untouched here.

## Checklist

- [x] Linked issue — closes #288 and #284
- [x] Updated docs, tests, and manual testing guidance — CHANGELOG, `docs/current-metrics.md` (unit/assertion/line-count rows and E2E 86 → 92), design brief, readmes
- [x] No sensitive details disclosed publicly

## Pre-existing issues found but deliberately not fixed here

1. `capture-screenshots.spec.ts` writes the **break-glass recovery notice to slot 9**, but the committed `screenshot-9.png` and readme caption 9 are the Users-list view, and caption 10 has no capture step — so `npm run screenshots` overwrites shot 9 with a different surface than the readme advertises. Documented in that file's header rather than silently renumbered; it needs a decision about which asset set is canonical.
2. `Admin_Bar::enqueue_assets()` hooks `admin_enqueue_scripts`, so `wp-sudo-admin-bar.js` **does run on `post.php` / `post-new.php`**, and at `r <= 0` it calls `window.location.reload()` — which will hit Gutenberg's unsaved-changes `beforeunload` prompt. Adjacent to this surface but out of scope.

https://claude.ai/code/session_01HZaaMe2j8NqwB9RRshaZdd


dknauss:
Follow-ups from the review, now filed and linked:

- **#300** — `Admin_Bar::enqueue_assets()` hooks `admin_enqueue_scripts`, so `wp-sudo-admin-bar.js` runs on `post.php` / `post-new.php` and calls `window.location.reload()` at `r <= 0`, straight into Gutenberg's unsaved-changes prompt. Adjacent to this surface, deliberately out of scope here.
- **#301** — `npm run screenshots` writes the break-glass recovery notice to slot 9, but caption 9 and the committed `screenshot-9.png` are the Users-list view, and caption 10 has no capture step. Documented as `KNOWN DRIFT` in the spec header by this PR rather than silently renumbered, since fixing it means deciding which asset set is canonical.

Also added `a374e30`, a docs-only commit recording entry **#38** in `docs/llm-lies-log.md`: while fixing the first review round I wrote that the in-editor indicator "shipped in 4.7.0". It has not shipped — #262 is still in `## Unreleased`, and 4.7.0 was the reauth modal, in-modal 2FA, the nonce fix, and demos/docs. The reviewer caught it pre-merge. Logged because the shape is distinct from the existing entries: a true fact about a nearby artifact (`screenshot-10` *did* ship in 4.7.0) was generalized to the artifact under discussion, in the edit least likely to be re-reviewed, and it would have reached user-facing release notes.

Note the "Markdown link and anchor check" failure is expected and self-resolving: `readme.md` points the two new screenshots at `raw.githubusercontent.com/.../main/...`, matching every existing screenshot row, so they 404 until this merges. It is not one of the required checks.

dknauss:
Both Codex threads applied in `8ac7056` — both were legitimate, neither was noise.

**P2 (POT).** Correct and a real gap: `Sudo · expiring` is the third accessible-name state added here, and it was absent from the committed template, so translators had no way to supply it. Regenerated with `composer i18n:make-pot`; `composer verify:i18n` now reports the template in sync. The only new `msgid` in the diff is that string.

**P1 (cite the Gutenberg DOM source).** Also correct — AGENTS.md requires the verification *source* in the commit message for third-party technical details, and the prior message asserted only that the selector was "verified against a live WP 7.0 editor," which is unauditable. I fetched the upstream files rather than restating what a live DOM inspection showed:

```
packages/interface/src/components/complementary-area-toggle/index.js
  L45  const identifier = identifierProp || `${ context.name }/${ name }`;
  L59  aria-controls={ identifier.replace( '/', ':' ) }
packages/interface/src/components/pinned-items/index.js
  L21-23  className={ clsx( className, 'interface-pinned-items' ) }
```

https://raw.githubusercontent.com/WordPress/gutenberg/trunk/packages/interface/src/components/complementary-area-toggle/index.js
https://raw.githubusercontent.com/WordPress/gutenberg/trunk/packages/interface/src/components/pinned-items/index.js

Both fetched 2026-07-26 from trunk, now recorded in the stylesheet header alongside the coupling note. With `registerPlugin( 'wp-sudo-session-indicator' )` supplying `context.name` and `PluginSidebar`'s `name` the same, the identifier is `wp-sudo-session-indicator/wp-sudo-session-indicator`, so `aria-controls` resolves to `wp-sudo-session-indicator:wp-sudo-session-indicator` — matching the live DOM. Worth recording alongside it: `.replace()` swaps only the **first** `/`, so the mapping holds only while neither name contains one.

Gates after both changes: `composer test:unit` 1151/3396, `lint`, `verify:metrics`, and `verify:i18n` all clean.

PR #344, opened 2026-07-26
ci: unbreak docs-lint on main (red since #338)

dknauss:
`docs-lint` has failed on `main` since #338 landed — runs [30211461238](https://github.com/dknauss/Sudo/actions/runs/30211461238) (d665e19) and [30212011322](https://github.com/dknauss/Sudo/actions/runs/30212011322) (a094d3f), after passing on 1788beb.

**Cause:** #338 added a dated provenance caveat to `docs/core-sudo-gate-implementation-spec.md` — "the signed-updates keys lapsed 2021-04-01 … verified against `wordpress-develop` trunk, 2026-07-26". Those dates are the *point* of the passage: it is a claim about upstream state at a moment in time, which is what the verification rules ask contributors to write. But docs-lint flags any `YYYY-MM-DD` outside its canonical-docs allowlist.

**Fix:** put the spec on that allowlist, beside `core-action-gate-proposal.md`, which is already excluded for the same reason — both are dated analyses of external state, not prose that goes stale silently.

docs-lint is not a required check, so nothing was blocked. That is precisely why it sat red for two commits with nobody noticing, and it is the third instance today of a check that had quietly stopped meaning anything — after `e2e-visual.yml`'s swallowed exit code (#342) and `composer verify:i18n` being wired into no workflow at all (#332).

https://claude.ai/code/session_01V4kcGFoqkxivzk9BSRUhEs

### DIFF
--- a/.github/workflows/docs-lint.yml
+++ b/.github/workflows/docs-lint.yml
@@ -27,6 +27,7 @@ jobs:
             "docs/abilities-api-assessment.md"
             "docs/connectors-api-reference.md"
             "docs/core-action-gate-proposal.md"
+            "docs/core-sudo-gate-implementation-spec.md"
             "docs/current-metrics.md"
             "docs/collaboration-analysis.md"
             "docs/sudo-design-notes.md"


dknauss:
Superseded by #346, which landed first and fixed this the better way.

I proposed excluding `docs/core-sudo-gate-implementation-spec.md` from docs-lint's date scan. #346 instead rewrote the sentence to `verified against \`wordpress-develop\` trunk; see `docs/llm-lies-log.md` #39 for the dated citations` — keeping the provenance and moving the dates to the file that is *already* the canonical home for dated upstream claims.

That is better than my version on both counts: it does not widen a lint exclusion (mine would have stopped flagging any future stale date in that spec), and it points the reader at richer evidence than a bare date. `docs-lint` is green on `main` as of c3e0b3f and 0ee1191.

Closing unmerged. The finding behind it still stands and is recorded in #345: docs-lint sat red on `main` for two commits because it is not a required check — the third instance today of a check that had quietly stopped meaning anything.

Issue #320, opened 2026-07-26
core-gate: split the proposal — v1 recent-auth primitive vs. a separate provenance/policy project (landability)


dknauss:
**Strategic (Codex).** The proposal is now free of architectural digression, but its "minimal closure" still includes official installs, updates, language packs, activation, deletion, theme switching, connector credentials, and origin settings — and that breadth is what generates most of #302 (core update), #306 (App-Password/automation block), and #307 (auto-updater provenance). For core landability, split the proposed work:

1. **v1:** the recent-auth primitive + **browser/cookie-auth direct-code effects** + the necessary identity pivots.
2. **Separate project:** the **policy/provenance** work for automatic updates, API credentials, CLI, cron, and generic programmatic callers.

This preserves the security thesis while avoiding a first patch whose compatibility surface is *effectively all WordPress automation*. Decision for the maintainer; would reshape §8/§9 and the actor-class policy scope.

_Source: Codex full design review of `origin/main`@`4ba9a6f` (2026-07-26); repo claims verified against the PoC/spec. WP Sudo unit suite green (1,151 tests / 3,396 assertions)._

dknauss:
## Track-B decision anchor — core-gate proposal ordering (Claude + Fable/Opus + Codex, reconciled)

**This issue is the fork; decide it first.** It reshapes the rest.

**Order once #320 is decided:**
1. **Mechanism invariants (blockers):** #315 (ambient-bearer / auto-replay), #303 + #316 (observational hooks can't veto — *same* fundamental problem), #308 (flat-freshness cross-authorization — a mechanism decision, not just landability), #310 (AUTH_SALT-in-options → assurance fails against its own motivating primitive), #319 (atomic proof issuance — an acceptance prerequisite).
2. **Selected v1 closure:** #302 (core update **stays in v1** — a browser-initiated direct-code effect), #304, and the **blocking** #311 splits (App-Password shared sink, option-write bypasses, pluggable session managers, fail-closed recovery).
3. **Compatibility / adoption:** #305, #309, remaining #311.
4. **If #320 approves the split:** move #306, #307 into the separate provenance/automation project.

The invariant that ties #315↔#322 (plugin) together: *no proof may authorize an attacker-selected transaction without browser binding + action/target binding + deliberate confirmation.* The plugin (Strict cookies) proves the UX/token model; core (Lax + shared nonces) needs its own threat analysis.

dknauss:
[...]

Issue #319, opened 2026-07-26
core-gate: proof issuance is not atomic with cookie delivery, and misses cookie paths


dknauss:
**Severity: MEDIUM** (Codex). The PoC writes the server-side proof hash, then `if ( ! headers_sent() )` guards the `setcookie()`, **ignores `setcookie()`'s return**, and **`return true` unconditionally** (verified, PoC L61–L87). If headers were already sent, or cookie emission fails, the server has **invalidated any prior proof but the browser never receives the new one** — reauth is reported successful and then **immediately fails its own gate**: an unrecoverable reauth/replay loop until expiry/cleanup.

It also iterates only `array( COOKIEPATH, ADMIN_COOKIE_PATH )` (verified, PoC L76/L161), whereas core auth cookies use **`PLUGINS_COOKIE_PATH`, `ADMIN_COOKIE_PATH`, `COOKIEPATH`, and sometimes `SITECOOKIEPATH`** — matters because the proposal promises third-party routes are covered at the sink.

**Fix:** make proof creation **fail before changing server state** when headers cannot be emitted; check cookie-write results and **roll back** the new record on server-detectable failure; define **one canonical cookie-path/domain/security policy** mirroring the applicable auth context. Test under custom `COOKIE*` constants and TLS-offload setups.

Refs: [PoC issuance](https://github.com/dknauss/Sudo/blob/4ba9a6fafbbebb1b555bbc233db85da8dbc3c69b/docs/core-sudo-gate-poc-patches.md#L61-L87), [core cookie paths](https://github.com/WordPress/wordpress-develop/blob/07b1f8b1d25db182d1ac4c2529d97e3d0cb04aea/src/wp-includes/pluggable.php#L1193-L1198).

_Source: Codex full design review of `origin/main`@`4ba9a6f` (2026-07-26); repo claims verified against the PoC/spec. WP Sudo unit suite green (1,151 tests / 3,396 assertions)._

dknauss:
[...]

Issue #316, opened 2026-07-26
core-gate: the self-email gate is specified on personal_options_update, an observational do_action that cannot veto


dknauss:
**Severity: HIGH** (Codex). spec §4.1 (`core/change-own-email`) routes enforcement to **`personal_options_update`** ('with confirmation enabled, `send_confirmation_on_profile_email()` … so the chokepoint sees no change; detect at `personal_options_update` / the REST self-update'). But `personal_options_update` is a **`do_action`** hook — observational, its return is discarded — so it **cannot veto** the pending-email write. As specified, the self-email pivot (a Group-B credential route: email → password-reset takeover) has **no working enforcement mechanism** on the profile surface. Same class as #303 (do_action adapters can't block), but a **distinct path the spec explicitly names**, so tracked separately.

**Fix directions:** gate the self-email change at a **veto-capable** seam — e.g. the `personal_options_update`-adjacent write is too late, so intercept where the pending `_new_email` / confirmation is *initiated* (a filter that can short-circuit, or `wp_die`+stash on the profile POST handler before `send_confirmation_on_profile_email()` runs), and gate the REST self-update via its permission callback. Return the challengeable `WP_Error` / interstitial rather than observing the change after the fact.

---
_Source: Codex design review (2026-07-26), extending the Fable+Opus set (#302–#311). Verified against the spec._

dknauss:
**Codex full detail — plus a second (multisite ordering) problem.** Confirmed: `personal_options_update` is a `do_action` whose return is discarded; core's `send_confirmation_on_profile_email()` writes `_new_email`, sends the mail, and restores `$_POST['email']` before `edit_user()`, so a gate `WP_Error` there cannot produce the promised challenge/403 — without a blunt `wp_die()` the mutation continues. **Second problem:** on confirmation, **multisite updates the signup row directly before calling `wp_update_user()`**, so a downstream gate leaves **partial state** even if it stops the eventual email write.

**Recommendation (Codex):** gate at the explicit `user-edit.php?action=update` flow **before** `personal_options_update`, or refactor the confirmation callback into a **veto-capable helper**; **gate the confirmation endpoint too**; mark a pending `_new_email` record as **initiated under recent auth** (legacy/unmarked pending records require a fresh challenge at confirmation); perform all confirmation-side writes **only after the gate passes**. P1 — email→reset is a load-bearing closure pivot.

[...]

Issue #315, opened 2026-07-26
core-gate: auto-replay lets a copied session weaponize the victim's reauth to execute a pre-stashed code action


dknauss:
**Severity: HIGH** (Codex). spec §5.1 / §8 / acceptance-4 specify that **non-account** code actions (plugin/theme install/activate/switch/delete) are stashed and **replayed normally** after a successful challenge (only account/credential/principal mutations are non-replayable / reauth-then-resubmit). The stash is **per-user** (`≤5 stashes/user`, §5.1). 

**Attack:** an attacker holding a *copied* session (same user) triggers a gated **code** action (e.g. plugin install/activate) → it is stashed pending a challenge. The **victim**, in their own browser, then completes a reauth (for the attacker's stashed action, or their own) → the stashed **attacker** code action **auto-replays and executes**. The victim's legitimate proof-of-intent is turned into execution of the attacker's pre-stashed code — a confused-deputy hole in the shared per-user stash + auto-replay. This is distinct from #308 (flat-freshness reuse) and #303 (do_action adapters).

**Fix directions:** bind each stash to the specific session/proof (verifier) that created it, so a copied session's stash cannot be completed by the victim's challenge; and/or drop **auto-replay for code actions** — treat install/activate/switch/delete as **reauth-then-resubmit** like the account pivots (the safest option, since the whole point of gating them is deliberate intent); and/or bind the challenge to the exact stashed request hash so an unrelated proof cannot release it.

---
_Source: Codex design review (2026-07-26), extending the Fable+Opus set (#302–#311). Verified against the spec._

dknauss:
**Codex full detail — broaden this beyond auto-replay to "the proof cookie is an ambient bearer" (P1).** The root problem is bigger than the stash: the design keeps the auth-session token (to preserve nonces), so a **cookie clone shares that token and can mint/harvest valid nonces**. The PoC sets the proof cookie **`SameSite=Lax`** (verified, PoC L83 — comment: "Strict drops it on a top-level nav into wp-admin"). Lax is sent on a **top-level cross-site GET**. Two vectors:

[...]

PR #352, opened 2026-07-26
docs(core-gate): fix the signature contradiction, apply #320 throughout, quarantine the PoC

dknauss:
Consolidation pass over the core-gate proposal docs, from a review whose findings were independently verified against live sources before acting.

## 1. §10 asserted the opposite of §1 on core-update signatures

The proposal's §10 claimed core updates "are the exception: `Core_Upgrader` requests signature verification via `verify_file_signature()`". §1 of the same document said the opposite, correctly.

Verified before rewriting, reading the enclosing symbol rather than a grep hit: `Core_Upgrader::upgrade()` calls `$this->download_package( $current->packages->$to_download, false )` — `$check_signatures` is false, so core packages are never signature-checked.

This matters beyond a stray sentence: it is a re-emergence of **`llm-lies-log.md` #39**, logged the same day. A design-review subagent confabulated this exact claim, it was about to drive a "narrow v1 ALLOW to core-signed packages" spec decision, it was caught and corrected in §1 — and the §10 copy was left standing.

Per the Prose Discipline rules added in #332, §10 now points at §1 rather than carrying a second copy of the citation, and records that its earlier revision claimed the inverse.

## 2. #320 had only reached two §1 paragraphs

The v1/provenance split was decided and #320 closed, but six locations still made non-interactive actor-class policy normative v1 behaviour, against a single §1 line saying those surfaces were deferred.

**The stale side was the opposite of what the review assumed.** `git log -S` shows spec §1 ("App Passwords, CLI, cron, XML-RPC are explicitly deferred") landed in PR #260, and §9's "the block/allow decisions still enforce in v1" landed in PR #281 — a day *before* #320 existed. §1 was a leftover PR #281 superseded and forgot to delete. #320 then happened to land on §1's posture, so §1 is kept as the single normative scope statement and the six sites reshaped to defer to it — per #320's own follow-on instruction to "reshape the actor-class policy scope to v1's narrowed surface".

Reshaped: spec §2 (threat model row), §3.6, §5.2, §9, §10 AC7; proposal §4, §5.

Two judgment calls worth reviewing:

- The earlier "excluding non-interactive routes from enforcement is a category error" argument is **retained and explicitly deferred**, not deleted. It is still sound, and it is why #306/#307 exist. The docs now say the category error is deferred, not resolved.
- **AC7 is inverted.** v1 cannot assert anything about non-interactive callers, so the acceptance criterion now tests the *misclassification boundary* — that no cookie-authenticated REST call is ever resolved as an API-credential caller.

The residual risk (REST-over-Application-Password install stays open in v1) is now stated plainly rather than left to inference.

## 3. The PoC patches doc is quarantined

It warned only that it was "illustrative and untested" while containing eight known-invalid designs, and closed with "Why this is the whole thing". It now opens with a SUPERSEDED / DO-NOT-IMPLEMENT banner mapping each defect to its issue (#302, #303, #306, #307, #308, #315, #319), including the missing `update_core()` seam that #320 explicitly kept **in** v1.

The closing section is marked superseded rather than deleted — the chokepoint-insertion argument still holds even though the security design carrying it does not.

One correction to the review that prompted this: the stash sketch has **no ownership binding at all** (the key rides a URL query arg), not the per-user keying that was alleged. It is weaker than described.

## 4. The proposal had no known-gaps section

`core-action-gate-proposal.md` is the document aimed at core contributors, and it carried §12's P1 findings with no caveat and exactly one issue reference in the whole file. Added a top-of-document callout plus inline markers at §4 ("the design that survives", #315) and the §3 Group-B email row (#316).

## Checks

- `docs-lint` passes — the spec and PoC are not on the date/count exclusion list
- `composer verify:sources` passes — 9 citations checked, all snippets present, no new orphan citations
- `composer verify:metrics` in sync

## Follow-up not done here

`docs/core-sudo-gate-implementation-spec.md` is on the grandfathered-orphans list in `bin/verify-sources.sh` ("migrate as you touch it"). This PR touches it but does not migrate its citation set to the registry. That is a substantial separate pass, and it is blocked on a design question: the registry's `GB-` prefix currently admits only Gutenberg sources, so registering `wordpress-develop` claims needs the prefix check and the dangling-reference scan widened together.


dknauss:
## Codex review triage

One verify-and-triage pass over the five threads. Four were valid and are fixed; one was already resolved by an amend that landed after the review ran.

**§5.4 still governed by actor-class policy (P1) — already fixed, thread is stale.** The review ran against `2c44bb9`; the amend at `68c5ca1` had already rewritten §5.4 to "out of v1 scope entirely (§1) and pass through ungated," with the actor-class policy explicitly labelled the follow-on project's design.

**Register the WordPress-core claim (P1) — valid, fixed.** Added `GB-CORE-NOSIG` to `docs/upstream-sources.md`, verified against live trunk (`Core_Upgrader::upgrade()`, L128, `download_package( …, false )`), and the §10 prose now references the ID instead of restating the mechanism. `composer verify:sources` → 10 citations, all snippets present.

Noting the constraint for the record: `bin/verify-sources.sh` hard-enforces a `GB-` prefix and its dangling-reference scan only greps `GB-*`, so a `wordpress-develop` source has to borrow the Gutenberg prefix today. The row says so. Widening the prefix + scan together is filed separately rather than smuggled into a docs PR.

**"Programmatic" is not an actor class (P1) — valid, and the most useful finding here.** This was a real security error, not a wording problem: classifying by *call origin* would let any hijacked session bypass the gate by invoking the sink from a plugin hook instead of posting the form. Removed "generic programmatic callers" from the deferral list in both docs and stated the rule — **actor class is a property of the request, never of the call site**; a nested programmatic call inherits the request's class. The exemption now covers only genuinely actorless contexts (cron, CLI) and non-cookie credential contexts (App Password, XML-RPC).

Specifying the trusted classification signal, and proving no cookie-authenticated request can resolve as actorless, is now #357 with §10 AC7 as its acceptance criterion.

**Stash ownership claim (P2) — valid, and my correction was the wrong correction.** Confirmed against `includes/class-request-stash.php`: `save()` stores `user_id` (L148) and `get()` rejects a mismatch (L202), so the ported implementation does have per-user ownership. The banner row now states the actual defect — ownership is **per-user**, which is the wrong granularity: not bound to the browser that answered the challenge, nor to the action and target, which is what lets a victim's reauth release an attacker-planted stash (#315).

**Mark superseded in the docs index (P1) — valid, fixed.** `docs/README.md` sent readers to the sketch under "What do the actual patches look like at the chokepoints?" with no warning. Retitled and flagged; the row now reads as an early sketch, superseded and not to be implemented.

All gates green after the amend: `verify:sources` (10 citations), `verify:metrics` in sync, docs-lint pass.

PR #364, opened 2026-07-26
chore(release): bump main to 5.0.0 (take the MAJOR), record the development lanes, correct the wp.org posture

dknauss:
Two commits, both making canonical release docs tell the truth.

## 1. `chore(release)`: bump `main` to 4.9.0 + record the lanes

`main` carried the released `4.8.0` identity in all five version-sync points, so an unreleased tree was indistinguishable from the shipped plugin and the next release boundary had no marker.

- `WP_SUDO_VERSION` → `4.9.0` in its four points (`wp-sudo.php` header + constant, `tests/bootstrap.php`, `phpstan-bootstrap.php`) and the `readme.txt` Stable tag, keeping all five in sync per `VERSIONING.md`.
- `docs/release-status.md` records the development-lane split.

**`blueprint.json` deliberately untouched** — still the `v4.8.0` tag ZIP. `VERSIONING.md` requires the Playground target to be bumped *after* the tag is cut; the public badge loads it from `main`, so a pre-tag bump fetches a nonexistent ZIP. Tag-time step, not drift.

**`4.9.0`, not `4.9.0-dev`.** No `-dev` suffix has ever been used here and `VERSIONING.md` defines no such convention; a `-dev` Stable tag would also desync the five points that rule keeps together. #348 already documents its proof format as `@since 4.9.0`.

**Upgrade behaviour: no routine runs.** The highest `Upgrader::UPGRADES` entry is `4.0.0`, so a site stored at `4.8.0` skips every routine and just advances the stored version — one option write on the next request.

### Lanes recorded

| Lane | Milestone | Clock |
|---|---|---|
| Plugin security release | `4.9.0 — security` | **Yes — the only one** |
| Plugin, everything else | `post-4.9.0` | No |
| Plugin, next major | `v5.0.0` | No |
| Core proposal research | `core-gate proposal — v1 readiness` | No |
| Core proposal, deferred | `core-gate: provenance/automation policy (deferred project)` | No |

All open issues are milestoned; none orphaned.

## 2. `docs`: wp.org submission is not planned, PCP is not a gate

The docs said submission was "intentionally delayed/on hold, but the repository should remain submission-ready at any time". That overstated it — submission is **not planned**, and Plugin Check is used only for occasional local testing. It is not among `main`'s six required checks, so a PCP finding does not block a release.

[...]

Issue #322, opened 2026-07-26
Security (P1): stash auto-resume is a confused deputy — a planted stash + victim reauth executes the attacker's transaction


dknauss:
**Severity: P1.** Verified in the shipping code (Codex finding, independently confirmed). Distinct from #315 (which frames this against the *core-gate proposal/PoC*); this is the **live plugin**.

## The flaw
The request stash is bound to **`user_id` only, not the originating browser**:
- `save()` stores `'user_id'` and nothing browser-distinguishing — [class-request-stash.php:148](includes/class-request-stash.php#L148); `get()` ownership check compares `user_id` alone — [:201](includes/class-request-stash.php#L201). The only per-stash secret is the random key, which travels in the challenge URL.

So a **confused-deputy** sequence works:
1. A cloned authenticated session (stolen cookie, no password) triggers a gated request → a stash is created, keyed to `user_id`.
2. The attacker sends the victim the challenge URL (they know the stash key).
3. The victim reauthenticates → Sudo **immediately replays the attacker's stashed request** — [class-challenge.php:661](includes/class-challenge.php#L661); if sudo is already active, `render_resume_page()` auto-resumes the stash with no fresh confirmation — [:944](includes/class-challenge.php#L944).

## Why it's more than harmless form-restore
Replay is enabled for **principal, credential, and critical-setting** actions, not just benign form data. The **`auth.app_password`** rule allowlists **`success_url`** in its stash — [class-action-registry.php:484](includes/class-action-registry.php#L484) — so a planted, replayed Application-Password authorization is a plausible **mint-a-durable-credential-and-exfil-to-attacker-URL** route after the victim's reauth.

## Scope / what it is *not*
- **Not** an unattended "stolen cookie instantly = RCE" bypass: it needs the attacker to induce the victim to visit the challenge URL and reauthenticate within the ~5-min stash TTL. But it **breaks the claim that reauthentication proves the victim intended the particular transaction.**
- Sudo's session cookie is **`SameSite=Strict`** — [class-sudo-session.php:891](includes/class-sudo-session.php#L891) — so #315's *direct* cross-site top-level-GET nonce-harvest variant does **not** transfer (Strict is withheld on cross-site navs). The **planted-stash** vector above still applies (victim navigates on-site).
[...]

Issue #280, opened 2026-07-26
Reauth lockout has no out-of-band clear — weaponizable to block sudo-gated remediation


dknauss:
## Summary

An active reauth lockout (`_wp_sudo_lockout_until`, set after 5 failed attempts, `LOCKOUT_DURATION = 300s`) has **no out-of-band clear**. Verified: `reset_failed_attempts()` (`includes/class-sudo-session.php:1210`) runs only on a **successful** reauth (impossible while locked — `attempt_activation()` returns `locked_out` before the password check, `:426-438`) or on natural 5-minute expiry (`is_locked_out()`, `:664-678`). There is **no WP-CLI command, no constant, and no admin action** that clears a lockout.

## Attack

A party holding the victim's session (stolen-cookie replay, or an active same-origin XSS running from the victim's own IP so the per-`(ip, user)` scoping at `:1125-1144` does not help) can submit 5 wrong passwords to the reauth AJAX endpoint and lock the victim out. While locked out the victim cannot obtain a **new** sudo session, so cannot perform any sudo-gated remediation (deactivate a malicious plugin, revoke sessions, change their own password) for the 5-minute window — and the block can be re-triggered indefinitely. The attacker sacrifices nothing: they cannot reauth anyway (no password).

The only escapes are host-level (FTP, or `wp plugin deactivate` if the WP-CLI surface policy allows) — nothing inside wp-admin.

## Scope note

The lockout gates reauth **attempts**, so it only bites when the victim has no active sudo session and needs a fresh one; an already-active session can still act until it expires. Bounded (5-min windows) but asymmetric and with no in-product recovery.

## Suggested fix

1. Add an out-of-band clear: a WP-CLI `wp sudo unlock --user=<id|login>` command, and/or a break-glass constant, so an operator with shell/config access can lift a lockout without waiting.
2. Consider preferring **escalating delay** over a **terminal block** for the *action-gate* decision (distinct from the auth-attempt decision) so a locked account can still be reached for remediation, per the red-team's recommendation.

## Source

Design red-team against WordPress 7.0.2, verified against the lockout code paths. See `.planning/reconciliation-carryover.md` (Part B.7 / D1-10).

_Filed from the Codex+Fable core-gate review carry-over (Part B)._

Issue #279, opened 2026-07-26
Per-user sudo proof key: a second browser's activation silently invalidates the first


dknauss:
## Summary

The sudo proof token is stored under a single per-user meta key, `TOKEN_META_KEY = '_wp_sudo_token'` (`includes/class-sudo-session.php:43`). `set_token()` (`:851-912`) does `update_user_meta($user_id, TOKEN_META_KEY, hash('sha256', $token))`, overwriting any previous value for that user.

So when a user activates sudo in **browser B** while they already have an active sudo session in **browser A**, B's activation overwrites the stored hash. A's `wp_sudo_token` cookie no longer matches, so `verify_token()` (`:920-961`) fails for A and A's sudo silently ends. A user can effectively hold only **one** active sudo browser at a time.

## Why it matters

Mostly a correctness/UX limitation (concurrent admin sessions are legitimate — desktop + laptop, or multiple tabs across profiles), with a subtle security edge: sudo state is keyed to the user, not to the specific login session that earned it. WordPress supports many concurrent sessions per user; the sudo proof should too.

## Suggested fix

Key the proof per **login-session token verifier** rather than per user — e.g. store the proof hash keyed by `wp_get_session_token()` (hashed), so each browser's sudo is independent and one browser reauthenticating does not revoke another's. This also aligns with reading the session store on the enforcement path (see the `destroy_all()` one-request-lag note, `class-sudo-session.php:52-57`).

## Source

Design red-team against WordPress 7.0.2. See `.planning/reconciliation-carryover.md` (Part B.2).

_Filed from the Codex+Fable core-gate review carry-over (Part B)._

PR #372, opened 2026-07-26
chore(release): bump main to 5.0.0 (take the MAJOR), record the development lanes, correct the wp.org posture

dknauss:
Supersedes #364 — same branch, same commits (`eb6894f`), renamed from `chore/bump-main-4.9.0`. GitHub closes a PR when its head branch is renamed, so this is a fresh PR rather than a retarget. No code differs.

Three commits, all making canonical release state tell the truth.

## 1. Bump `main` off the released identity

`main` carried the released `4.8.0` identity in all five version-sync points, so an unreleased tree was indistinguishable from the shipped plugin and the next release boundary had no marker.

**`blueprint.json` deliberately untouched** — still the `v4.8.0` tag ZIP. `VERSIONING.md` requires the Playground target to be bumped *after* the tag is cut; the public badge loads it from `main`, so a pre-tag bump fetches a nonexistent ZIP.

**Upgrade behaviour: no routine runs.** The highest `Upgrader::UPGRADES` entry is `4.0.0`, so an upgrading site skips every routine and just advances the stored version.

## 2. Correct the WordPress.org posture

The docs said submission was "intentionally delayed/on hold, but the repository should remain submission-ready at any time". Submission is **not planned**, and Plugin Check is local-only — not among `main`'s six required checks. This mattered concretely: the old text made `readme.txt` Stable tag look like it carried a wp.org contract (selecting a served SVN tag) when it is only zip metadata.

Historical records (`CHANGELOG.md`, `.planning/MILESTONES.md`, `RETROSPECTIVE.md`, dated per-release sections) are **deliberately unchanged** — they record what was true then.

## 3. Take the MAJOR: 5.0.0

The `#322` fix (#368) removes the documented `wp_sudo_action_replayed` hook (event code `1900009`). `VERSIONING.md` classes removing a documented hook as **MAJOR**. Taken by the rule, not by override. Every other payload item (#278, #279, #280) would have been MINOR alone.

Beyond version strings, this fixes the `@since` / `pre-X` markers #348 merged into `main` while the target was undecided (`class-sudo-session.php`, `class-site-health.php`, `uninstall.php`, tests, E2E, `security-model.md`), and regenerates `languages/wp-sudo.pot` — its `Project-Id-Version` header carries the plugin version and was **still reading 4.8.0**, so the first bump had missed it. Caught by `verify:i18n`.

`release-status.md` also loses two claims the decision falsified: that the next tag would "most likely" be MINOR, and that the call was open.

## Lanes recorded

| Lane | Milestone | Clock |
|---|---|---|
| Plugin security release | `5.0.0 — security` | **Yes — the only one** |
| Plugin, next feature batch | `v5.1.0` | No |
| Plugin, everything else | `post-5.0.0` | No |
| Core proposal research | `core-gate proposal — v1 readiness` | No |
| Core proposal, deferred | `core-gate: provenance/automation policy (deferred project)` | No |

**Milestone collision resolved:** a `v5.0.0` milestone already existed for a curated feature batch. Its own description called the number a product-signaling call, and all five issues in it are backward-compatible additions — renumbered to **`v5.1.0`**, where MINOR is the correct class.

## Verification

`composer test:unit` (1195 tests, 3501 assertions), `lint` (24/24), `analyse:phpstan`, `analyse:psalm`, `verify:metrics`, `verify:i18n`, and the `docs-lint` date/count check run locally.

⚠️ **The CLAUDE.md pre-commit reviewer did not run** — no `reviewer` subagent type is registered in this session and no pre-commit hook is installed. I did not write the `reviewer-approved` flag. Treat the PR checks as the gate, or run the reviewer before merging.


dknauss:
Both P2 findings verified against the files and applied in `f578dec`.

**ROADMAP** — confirmed: `applies_to: 4.8.x`, "the **v5.0.0** milestone", and #219/#239/#240 all labelled `v5.0.0`, which now collides with the security release. Retargeted to `v5.1.0` (milestone number unchanged, so the links still resolve) and `applies_to` moved to `5.0.x`.

**release-environment-log** — confirmed: the introduction and the standing blocker-posture rule both said submission awaited maintainer approval. Both updated. Per your note, the dated per-release rows are left alone — they record what was true at each tag.

Worth recording why the second one was missed: my earlier wp.org sweep grepped for `wporg|WordPress.org|wp.org|submission-ready|Plugin Check|PCP`, but that invocation died on a zsh glob error and the narrower retry never covered a file phrasing the policy as "awaits publication approval" without any of those tokens. The file read as clean when it wasn't. Good catch.

dknauss:
Follow-up: filed #392 against this PR's version decision.

Short version — the MAJOR was taken faithfully by the rule, but the premise was wrong about what #350 actually did. `#322` **moved** the `wp_sudo_action_replayed` call site rather than removing it; it still fires at `includes/class-challenge.php:1229`, in the bound-replay branch. #350 is "never *auto*-replay", not "never replay".

Checked the rest of the `v4.8.0..main` payload against `VERSIONING.md` too — no documented hook or public method removed (`verify_token()` was `private static`), min WP/PHP unchanged, slug unchanged. So no MAJOR trigger fired and `4.9.0` is the correct number.

No `v5.0.0` tag was cut, so this is `main`-only. Full evidence, commands, and the file-by-file scope are in the issue. Not touching release files myself since this lane owns them and is live — flagging rather than colliding.

Issue #360, opened 2026-07-26
core-gate: PoC vertical slices — A (code effects, landed, no core patch) / B (identity pivots, needs a wordpress-develop branch)


dknauss:
`docs/core-sudo-gate-poc-patches.md` is now banner-quarantined (#352) as a superseded, known-vulnerable sketch. This issue is its replacement.

## Why the sketch failed

It was pseudocode maintained *in parallel* with an evolving spec, so it drifted silently. By the time it was audited it carried nine defects — `SameSite=Lax` ambient proof cookie, server-side stash + auto-replay, non-vetoing pre-op hooks, non-atomic proof issuance with an unchecked `setcookie()` return, incomplete cookie paths, per-user-only stash ownership, non-interactive policy that #320 had moved out of v1, flat-freshness v1 callers, and a **missing `update_core()` seam** that #320 explicitly kept *in* v1 (#302).

None of those were caught by anything, because pseudocode has no CI. The lesson is not "write a better sketch."

## What to build instead

One **vertical slice** as a real patch branch against `wordpress-develop`, with tests that run:

- One browser-driven code effect (`activate_plugin()` is the cheapest real one)
- One identity pivot (own-password change — the #20140 case)
- Browser-bound proof, per-session, not per-user
- Explicit action/target confirmation — **no auto-replay** (#315)
- Cookie-authenticated REST returning the challenge shape, not a silent 200
- API-credential and CLI/cron callers out of scope per #320 — verify they pass through unchanged
- The `update_core()` seam (#302), since it is v1

## The test that matters most

**Two real browsers, cloned session.** Browser A answers the challenge; Browser B holds a copy of the same auth cookie and must *not* inherit the elevation. That single test is what falsifies or confirms the proposal's central claim, and nothing in the current artifact set exercises it.

Also worth a test: the misclassification boundary from #357 — a programmatic call inside a cookie-authenticated request must be challenged, not exempted.

## The split (updated — Slice A has landed)

The original framing assumed the whole PoC needed a `wordpress-develop` patch. That was written before we verified core's existing seams, and it is only half true.

### Slice A — code effects. **Done, and needed no core patch** (#380)

`WP_Upgrader::install_package()` already carries a vetoing short-circuit:

```php
$res = apply_filters( 'upgrader_pre_install', true, $args['hook_extra'] );
if ( is_wp_error( $res ) ) {
    return $res;
}
```

[...]

Issue #358, opened 2026-07-26
core-gate: audit every chokepoint for whether it can actually veto (umbrella for #303/#316)


dknauss:
Umbrella for the single design question underneath several separately-filed P1s.

## The observation

#303 (pre-op adapters are `do_action()` hooks that cannot block) and #316 (the self-email gate is on `personal_options_update`, also observational) are filed as separate findings. They are the same problem: **the spec picked seams that physically cannot refuse.**

This is also the exact defect §4.1.1 uses to *reject* the Abilities API as an enforcement point — `WP_Ability::execute()` fires `wp_before_execute_ability` and calls the ability on the next line, discarding the return. The spec then prescribes the same pattern at `delete_user`, `grant_super_admin`, `switch_theme`, and `personal_options_update`. That is an internal-consistency failure, not just a gap, and it is the thing most likely to sink the proposal in a core review.

The proposal claims a "provable minimal closure." If several sinks in that closure cannot block, the closure is not proven.

## What this issue tracks

A per-chokepoint table covering **every** entry in the §4.1 catalog, with one row each:

| Column | Meaning |
|---|---|
| Sink | The function or hook |
| Return contract | `WP_Error` / `bool` / `void` / bare `do_action` |
| Can it refuse today? | yes / no |
| Seam mechanism | return an error, pre-op adapter, permission callback, **new core filter**, or signature change |
| Cost to core | back-compat surface, whether it needs a `@since`, whether third-party callers break |

Known-hard rows to resolve, at minimum:

- `wp_delete_user()` — returns `bool`; a returned `WP_Error` is truthy, so the naive guard *inverts* the result
- `switch_theme()` — returns `void`
- `grant_super_admin()` / `add_user_to_blog()` — `do_action`
- `personal_options_update` / `edit_user_profile_update` — `do_action` (#316); note the REST self-update path may have a usable permission callback while the profile screen does not
- `wp_set_password()` — bypasses `wp_update_user()` entirely, and gating it breaks the unauthenticated reset flow (#304)

## Why it blocks other work

- **#315's fix depends on it.** Replacing auto-replay with a confirmation page needs a seam that can hold the operation *before* it commits.
- **The PoC rebuild depends on it.** There is no point generating a tested `wordpress-develop` patch until the seam per sink is settled.
- **AC1 depends on it.** "A stolen-cookie session cannot perform any initial-catalog action" is unprovable while some catalog actions have no blocking seam.

## Acceptance

[...]

Issue #357, opened 2026-07-26
core-gate: "programmatic" is not an actor class — specify the trusted classification signal


dknauss:
Found in review of #352.

## The problem

The v1 scope statement (spec §1) defers non-interactive callers. An early draft listed "generic programmatic callers" among them. That is a category error with a security consequence:

**A plugin calling `activate_plugin()` or `WP_Upgrader::install_package()` from inside an admin-screen load, or during a cookie-authenticated REST request, is an interactive actor.** The request has a live cookie session behind it. There is no "programmatic actor" — there is only the actor the *request* authenticated as.

If the gate classified by **call origin** ("this reached the sink from PHP rather than a form handler"), any hijacked session gets a one-line bypass: don't post the form, call the sink from a plugin hook. That would be a worse hole than the one the gate closes.

Conversely, if the gate classifies by the request's cookie actor (correct), then the claim "generic programmatic callers pass through unchanged" is **false** — a programmatic call inside an admin request gets challenged like any other interactive action.

Either way the original phrasing was wrong. #352 removes it and states the rule: **classification is a property of the request, not of the call site**; a nested programmatic call inherits the request's actor class.

## What this issue tracks

The rule is now stated, but the *mechanism* is not specified:

1. **Name the trusted classification signal.** What does the gate read to resolve actor class — and can it be spoofed by anything reachable from a compromised session? Candidates: presence of an auth cookie / `wp_get_session_token()`, `WP_CLI` constant, `wp_doing_cron()`, the REST authentication callback's result. Each needs a "can an in-request attacker influence this?" answer.
2. **Prove the actorless branch is unreachable from a cookie request.** The dangerous direction is a cookie-authenticated request resolving as actorless/CLI and skipping the gate. `wp_doing_cron()` is true for a request to `wp-cron.php`, which is **unauthenticated-triggerable** — so "cron" must not be inferrable in a way that a browser request can induce.
3. **Decide the WP-CLI + cookie edge.** WP-CLI with `--user=` sets a current user without a session token. It is out of v1 scope, but the classifier must not mistake it for a cookie session, nor a cookie session for it.
4. **Write the AC.** §10 AC7 is currently the acceptance criterion ("no interactive-class request is misclassified as non-interactive"); it needs concrete test cases, at minimum: a cookie-authenticated REST call, a programmatic `activate_plugin()` inside an admin request, and a real `wp-cron.php` hit.

## Not in scope

[...]

Issue #307, opened 2026-07-26
core-gate: auto-updater ALLOW branch depends on a provenance primitive core lacks


dknauss:
**Severity: HIGH** (Fable). §9's one ALLOW branch — 'package from the site's configured update source' — rests on a host-allowlist / pinned-origin concept **WordPress does not have**. `install_package()` receives only an unpacked local path, so provenance must be set at the filterable `upgrader_pre_download` or derived from the mutable `update_plugins`/`update_core` transients. Whatever can influence that hook or transient can mark an arbitrary package **trusted** and reach `install_package()` on the ALLOW branch with no challenge and no actor — and `wp-cron.php` is unauthenticated-triggerable. `wp_action_gate_trusted_autoupdate()` is an undefined predicate carrying the whole weight of the one whitelist. **Fix:** a concrete non-filterable definition (flag set only inside `WP_Automatic_Updater` for a package whose URL host matches the api.wordpress.org offer fetched in the same request, passed as a private arg), or narrow ALLOW to core-signed packages only.

---
_Source: Fable+Opus two-model design review of the core-gate proposal/spec (2026-07-26). Doc-anchored claims verified by grep; WP-internal claims to be re-checked against live core before landing._

dknauss:
**Scope depends on #320** (the proposal split decision). If v1 sheds the automation/provenance surface, this moves to the separate provenance/policy project — so it is deliberately **not** in the "v1 readiness" milestone and carries no priority yet. Hold until #320 resolves.

dknauss:
Moved out of v1 to milestone **core-gate: provenance/automation policy (deferred project)** per the #320 split decision. #307 *is* the provenance problem: §9's auto-updater ALLOW branch reintroduces the 'trusted origin' exemption key that proposal §3 rejects, on a primitive core lacks. Deferring it removes that §3-vs-§9 contradiction from v1; the origin-blind package-write gate stands on its own. Not started until v1 lands.

dknauss:
[...]

PR #380, opened 2026-07-26
poc(core-gate): the install_package() vertical slice, tested against real core

dknauss:
First slice of #360, replacing the pseudocode quarantined in #352.

## The finding this rests on

**Core already ships a vetoing seam at the package-write chokepoint.** `WP_Upgrader::install_package()` opens with:

```php
$res = apply_filters( 'upgrader_pre_install', true, $args['hook_extra'] );

if ( is_wp_error( $res ) ) {
    return $res;
}
```

Same short-circuit shape as `pre_delete_post` — the filter's error return aborts the operation *and becomes the return value*, so every caller that already handles `install_package()`'s `WP_Error` surfaces the refusal unchanged. Registered as `GB-UPGRADER-PRE-INSTALL`.

That means the **Group-A code-execution closure — the part carrying the "de-escalate every XSS from RCE" argument — needs no core change at all.**

## This changes #360's shape

#360's acceptance says "a branch that applies to `wordpress-develop` trunk." That was written before we knew this seam existed. It does not apply to this slice: there is nothing to patch. So #360 splits cleanly:

| Slice | Needs a core patch? | Status |
|---|---|---|
| **A — code effects** (`install_package`, `activate_plugin`, `wp_edit_theme_plugin_file`) | **No** — seams already exist | this PR |
| **B — identity pivots** (`wp_set_password`, `set_role`, `wp_delete_user`, `grant_super_admin`) | **Yes** — those sinks cannot refuse | blocked on #358's four `pre_*` filters |

Which is a much better story for a Make/Core post: *here is the code-execution gate working today against core's own seam; here are the four filters needed to finish it.*

## The control test is the point

A test asserting `install_package()` returned an error proves nothing on its own — it might have failed for unrelated reasons. So the suite runs the **identical call twice**, once with the filter attached and once without:

| | outcome | package written to disk |
|---|---|---|
| gate on | `wp_sudo_reauth_required` | **no** |
| gate off | success | **yes** |

That difference is the entire claim of proposal §3, and it is now empirical. I verified it by running it, not by reasoning about it.

## Coverage

8 tests, 23 assertions, against live WordPress + MySQL:

- interactive session without proof → refused, nothing written
- interactive session with proof → the gate is not what stops it
- **cloned session does not inherit the proof** — the §4 claim the proposal stands or falls on
- no session (cron / WP-CLI / API credential) → passes through, per the #320 v1 scope. This is the branch that keeps background security updates working
- programmatic call inside an interactive request → **still gated** (#357): call origin is not an actor class, or calling the sink from a plugin hook is a one-line bypass
- forged cookie cannot mint or hold a proof — `wp_get_session_token()` only *parses* the cookie, `WP_Session_Tokens::verify()` is the real check
- revoking re-gates immediately

## Two design facts worth carrying into the spec

**Provenance genuinely cannot be read at this seam.** `WP_Automatic_Updater` reuses `Plugin_Upgrader`/`Theme_Upgrader` and adds **no marker of its own** to `$hook_extra` (verified against trunk). Nothing in the payload distinguishes an auto-update from a manual one — which is exactly why #307 needs a provenance primitive decided upstream, and why v1 keys on *absence of a session* instead. That is a concrete confirmation of a claim the spec previously asserted.

**The proof is per-session, never per-user**, and the token is hashed because the key becomes an option name.

## Isolation and gates

Self-contained under `poc/` with its own bootstrap and phpunit config, so it cannot perturb the plugin's suites. `phpcs` and `phpstan` scope to production paths, so it sits outside them by design — this is proof-of-concept code, not shipped.

`docs/current-metrics.md` updated **first** for the repo-PHP-line row, per the metrics rule. All green: `verify:metrics`, `verify:sources` (28 citations), docs-lint, phpcs, psalm, and the plugin's own 1195-test suite.

## Not in this slice

No challenge UI (proposal §4 — tests grant the proof directly), no identity pivots (blocked on #358), no non-interactive policy (out of v1 scope per #320).

Refs #360, #358, #357, #320, #315, #307


dknauss:
## Codex triage — one pass, nine findings, essentially all valid

Two were real defects, one was a packaging leak, four were unregistered claims, two were silent omissions. Fixed in `1d3d1b1`.

**Bind the proof to a separate browser secret (P1) — correct, and it invalidated my headline claim.** `test_a_cloned_session_does_not_inherit_the_proof()` called `start_session()` twice, minting a *new* token for browser B. That proves isolation between two independent logins. A real cookie thief presents the victim's *same* token, so the token-keyed proof resolved to the same record and the copy inherited it — and the PR body called that test "the claim the proposal stands or falls on."

`grant_proof()` now issues a random secret in its own cookie and stores only its hash, per §4. The old test is renamed to what it actually proves; `test_a_copied_login_cookie_does_not_inherit_the_proof()` replays browser A's exact cookie. Mutation-checked: removing the secret check reds it.

**Reject invalidated sessions (P2) — correct, and I'd rate it higher than P2.** It was a fail-open. A cookie-bearing request whose token no longer verified became `ACTOR_NONE` and fell into the out-of-v1-scope branch, so the write was **allowed**. "Log out everywhere" un-gated instead of re-gating. Now `ACTOR_REVOKED`, refused; only a genuinely cookieless request passes. Mutation-checked.

**Exclude the PoC from the PressShip package (P2) — correct and material.** `.pressshipignore` did not exclude `poc/`, so it would have shipped to WordPress.org while the README said "not shipped". Your read of why `bin/build-plugin-check-dist.sh` was unaffected — allowlist rather than denylist — is right. Added `poc/`.

**Register the cookie-parser and auto-updater claims (P1 ×2) — correct.** Added `GB-AUTH-COOKIE-PARSE` and `GB-AUTOUPDATER-UPGRADE`, both verified against trunk. On the auto-updater: I've phrased it as a positive citation (it instantiates the ordinary upgraders and calls `upgrade()` with generic options) rather than asserting an absence, since "adds no marker" isn't something a `must-contain` snippet can verify.

**Verify the package-write closure / every caller propagates (P1 ×2) — correct.** Both were scope words the one row can't support. Narrowed: "single funnel" now states that the funnel claim is the proposal's and this slice proves a veto *at* the sink, not that every producer reaches it; the caller-rendering sentence is scoped to what the tests show.

**Metrics acknowledgement (P2) — correct.** Narrowed to production and plugin-test metrics, which `poc/` genuinely doesn't touch.

**Action-binding (#308) and proof signing (#310) — correct that they're absent; deferring rather than fixing.** Both are real requirements, both are tracked, and flat freshness is the spec's current v1 recommendation. What was wrong was omitting them *silently* — that is precisely how the previous PoC accumulated nine defects — so they're now in the "deliberately does not do" list with their issue numbers.

Gates: 10 tests / 34 assertions, `verify:sources` 31 citations, `verify:metrics` in sync, phpcs, psalm, and the plugin's 1227-test suite all green.

dknauss:
## Second triage round — both P1s valid, and one of them corrects the proposal

**Gate packages before extracting them into `wp-content` — correct, and this is the most useful thing the review has produced.**

Verified in `WP_Upgrader::run()`:

| step | line | effect |
|---|---|---|
| `download_package()` | 849 | fetches the archive |
| `unpack_package()` | 887 | **extracts into `wp-content/upgrade/`** (:367, :378) |
| `install_package()` | 898 | moves into the live tree — where the gate fired |

So the gate blocked the final move but not the extraction: attacker-controlled PHP was already in a directory many hosts execute PHP from. My integration helper called `install_package()` directly, which is exactly how the ordering escaped notice.

This is not just a PoC bug. **Proposal §3 names `install_package()` as *the* code-write seam, and on this evidence it is not sufficient on its own for the interactive branch.** I'll carry that into the spec separately. It is also the clearest argument yet for building slices instead of pseudocode — the sketch this replaced could never have surfaced an ordering fact.

Now gated at `upgrader_pre_download` as well, which fires before both the download and the unpack; the `install_package()` gate stays because that sink is reachable directly by callers that never go through `run()`. `test_the_run_path_is_refused_before_anything_is_unpacked()` asserts `wp-content/upgrade/` is untouched, and reds if the new filter is removed.

Two caveats I've kept in the README rather than let the fix imply more than it shows: the ordering was found by reading `run()`, not by demonstrating the race, and the residual depends on the host executing PHP under `wp-content`.

**Gate admin auth cookies when the logged-in cookie is absent — correct, and distinct from the earlier revocation fix.** `wp_get_session_token()` reads `LOGGED_IN_COOKIE` unconditionally while wp-admin authenticates from `AUTH_COOKIE`/`SECURE_AUTH_COOKIE`, so a client presenting a valid admin cookie and no logged-in cookie was fully authenticated with full capabilities, had no readable token, and got classified actorless. All three cookies are now checked.

I kept it as a cookie test rather than switching to `is_user_logged_in()` on purpose: WP-CLI `--user` and a cron callback calling `wp_set_current_user()` are authenticated without being browsers, and #320 keeps those out of v1 scope. Switching would have fixed this hole by creating a scope violation.

Both mutation-checked. 12 tests / 40 assertions; full gate set green after a rebase onto `15f7ecd`, with metrics recomputed rather than merged.

dknauss:
## Third triage round — three fixed here, the rest to #388. Closing the loop.

**Stop claiming flat freshness matches the current spec (P1) — correct, and the most useful catch of this round.** The spec moved during this branch's life. §5.1 now reads *"No reusable window for in-scope actions. A proof is minted for one action digest, is single-use, and is discarded on redemption"*, and §11 item 3 is **closed** — *"with no reusable window there is nothing to scope."*

So the slice's proof model is **superseded, not aligned**, and the README asserted the opposite. Rewritten to say so plainly, naming both gaps (per-action digest binding and single use; the §4.2 HMAC-signed record) and stating that the two seam findings are independent of the proof model and stand regardless.

This is the third time in this branch I've written a claim from the spec as I understood it rather than the spec as it is. Worth recording as a pattern, not just a fix.

**Keep the control call's destination identical (P2) — correct.** The blocked call targeted `$this->destination` and the ungated one a freshly generated path. Two different calls prove less than one call twice. Same destination now.

**Run the PoC suite in an automated validation lane (P2) — correct, and it matters more than P2 suggests.** The separate bootstrap and config are what stop the PoC perturbing the plugin's suites, and are also exactly why nothing ran it: every required gate could stay green while the load-bearing real-core tests were broken. Added `composer test:poc` plus a step in the integration job, which already installs the WordPress test library the bootstrap needs.

---

### Deferred to #388, with reasons

**The three "make the check detect X" findings are a genuine methodological point, and I'd rather fix it properly than patch it here.** A `must-contain` snippet verifies a *positive* fact. It cannot falsify an *absence* claim — if core added HMAC validation on the line after `explode()`, or an automatic-update marker in the arguments under `$upgrader->upgrade(`, both rows would still pass. That is true of `GB-CORE-NOSIG` on `main` as well, so it is a property of the checker, not of this branch. #388 proposes a `must_not_contain` column and lists the options.

**The two remaining row requests** (`pre_delete_post`'s contract; `wp_set_password()`/`set_role()` returning `void`) are in #388 too. The second is the one that actually matters — it is the premise of #358's "these sinks cannot refuse" argument and of the four-filter ask that follows.

**Proof-atomicity on grant (P2)** is #319's subject and applies to the real challenge handler, which this slice does not implement — `grant_proof()` exists so tests can mint a proof without one.

---

This is round three, and the substantive defects (copied-cookie inheritance, two fail-opens, the unpack ordering, the stale spec claim) are all fixed and mutation-checked. Remaining items are verification-method work on rows that mostly predate this branch. Landing here; #388 tracks the rest.

The current red is not this branch: the integration job hit `toomanyrequests: Rate exceeded` pulling `mysql:8.0`, and `PHPUnit` is a summary gate cascading from it.

PR #390, opened 2026-07-26
chore(release): roll main back to 4.9.0 — the MAJOR trigger no longer exists

dknauss:
## Why this isn't an override of the versioning rule

`5.0.0` was taken **by** the rule: #322 removed the only call site of `wp_sudo_action_replayed`, a documented public extension point (`developer-reference.md`, event code `1900009`), and [`VERSIONING.md`](VERSIONING.md) classes removing a documented hook as MAJOR.

**That fact no longer holds.** The #322 **v2** layer (PR #350) merged and restores replay — and with it the hook. `do_action( 'wp_sudo_action_replayed', … )` is live on `main` at `includes/class-challenge.php:1229`. Nothing documented is removed in the shipping state, so the MAJOR trigger is gone and the same rule now yields a **MINOR**.

This is the outcome `release-status.md` itself predicted when it deferred v2:

> *"it restores replay, and therefore probably restores the removed hook — so its own version classification must be decided on its merits, not inherited from this release."*

It was decided on its merits. The merged payload is additive plus security fixes → **4.9.0**.

## Changed (per the CLAUDE.md release checklist)
- `WP_SUDO_VERSION` in **all four** places + `readme.txt` Stable tag
- In-code proof-format boundary markers (`@since`, `legacy pre-X` comments in `class-sudo-session.php`, `uninstall.php`, `class-site-health.php`, and their tests) and the security-model mitigations heading — these name *the release the proof format changed in*, so they move with it
- `release-status.md`: the rationale is **rewritten, not deleted**, so the reversal and its reason stay on the record. The next feature batch returns to **`v4.10.0`** — it had only been pushed to `v5.1.0` because `5.0.0` was claimed.

## Deliberately untouched
`blueprint.json` still points at **`v4.8.0`**, the last tag that actually exists. Per `AGENTS.md` it is bumped **after** a tag is cut, never before — pointing it at an untagged version would serve a 404 to the public Playground demo.

## Gates
`test:unit` **1226 tests / 3611 assertions OK** · `lint` clean · PHPStan + Psalm clean · `verify:metrics` in sync.

### DIFF
--- a/docs/release-status.md
+++ b/docs/release-status.md
@@ -80,10 +80,10 @@ which is a ⚠️ **retroactive gap still open** for a maintainer reuse-or-rerun
 
 ## Current `main` release state
 
-- **Current `main` version:** `5.0.0` (runtime constant) — **unreleased**. The latest tag remains `v4.8.0` (cut 2026-07-23, on `10587a4`). `main` was bumped off the released identity on 2026-07-26 so that an unreleased tree stops advertising a shipped version; the bump carries no payload of its own. Unlike the `4.7.0` state, `main` has progressed **past** the tag with unreleased work (see next bullet). (`4.8.0` is the REST-gate security-hardening release plus the opt-in role/capability lockdown audit MVP; see the payload note above.)
+- **Current `main` version:** `4.9.0` (runtime constant) — **unreleased**. The latest tag remains `v4.8.0` (cut 2026-07-23, on `10587a4`). `main` was bumped off the released identity on 2026-07-26 so that an unreleased tree stops advertising a shipped version; the bump carries no payload of its own. Unlike the `4.7.0` state, `main` has progressed **past** the tag with unreleased work (see next bullet). (`4.8.0` is the REST-gate security-hardening release plus the opt-in role/capability lockdown audit MVP; see the payload note above.)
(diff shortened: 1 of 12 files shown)


dknauss:
Heads-up: #389, #390, and #391 are all doing this same rollback, opened within ~6 minutes of each other in response to #392. Comparison and a consolidation recommendation are in [#392](https://github.com/dknauss/Sudo/issues/392#issuecomment-latest) — short version, #389 covers all 15 files (the only one that also fixes `CHANGELOG.md` and `.planning/SESSION-HANDOFF.md`), so it is the one to land. Flagging so three lanes do not keep paying for the same change.

dknauss:
Closing as a duplicate — three sessions independently built this same rollback (#389, #390, #391). Keeping **#389**, which is the superset: it is the only one that updates `CHANGELOG.md` and adds the **Upgrade Notice for the forced reauth**, which matters because #322 forces a reauth on upgrade.

Everything this PR changed (the five version strings, the `@since 5.0.0` / `pre-5.0.0` references, `uninstall.php`, `wp-sudo.php`) is covered there. Nothing from here is being lost.

_🤖 Triaged by [Claude Code](https://claude.com/claude-code)_

dknauss:
Recommending this close in favour of **#389** once its rationale paragraph is grafted — I have posted the exact text on #389 and asked for it to be taken before this one closes.

The paragraph worth preserving is the one quoting this document's own earlier prediction, that v2 "restores replay, and therefore probably restores the removed hook — so its own version classification must be decided on its merits." None of the other rollback PRs has it, and it is what shows the `5.0.0` call was made in good faith and then overtaken by the exact event the doc anticipated.

#389 otherwise covers strictly more: 15 files against this PR's 12, including `CHANGELOG.md` (which still carries a `pre-5.0.0` string), `.planning/SESSION-HANDOFF.md`, and the `admin-bar-timer.spec.ts` reference. Comparison is in #392.

PR #391, opened 2026-07-26
chore(release): roll the unreleased version string back to 4.9.0

dknauss:
**Version string only.** #372's other content — the development-lane records and the wp.org posture corrections — is untouched.

Nothing has been released at either number: the newest tag is `v4.8.0`, and neither `v4.9.0` nor `v5.0.0` exists. This changes an unreleased identity, not a shipped one, so there is no upgrade path to worry about.

## Changed

All five sync points from the CLAUDE.md checklist, plus the POT:

| Location | |
|---|---|
| `wp-sudo.php` | plugin header `Version:` **and** `WP_SUDO_VERSION` |
| `phpstan-bootstrap.php` | `WP_SUDO_VERSION` |
| `tests/bootstrap.php` | `WP_SUDO_VERSION` |
| `readme.txt` | `Stable tag` |
| `languages/wp-sudo.pot` | `Project-Id-Version` |

Plus the in-tree references naming the **same** release, which would otherwise document a version that will never exist: `@since 5.0.0` and the `pre-5.0.0` legacy-key comments in `class-sudo-session.php`, `class-site-health.php`, `uninstall.php`; the "Since 5.0.0" / "(5.0.0)" notes in `docs/security-model.md`; and two test comments.

`docs/release-status.md`'s version **facts** are updated too — CLAUDE.md makes it the canonical source for release-state claims, so leaving them at 5.0.0 would have made it lie.

## Deliberately not changed

**Dependency versions.** `package-lock.json`, `composer.lock` and `.sbom/bom.json` contain `^5.0.0` constraints and `phpcsextra-1.5.0.0` — unrelated to this plugin. A blanket find-and-replace would have corrupted the lockfiles, which is the main hazard in a change like this.

**The "Why `5.0.0` and not `4.9.0`" rationale.** That is a decision record, and rewriting it is not a string change. It is marked superseded with a note stating the version is now 4.9.0, that the argument no longer describes `main`, and that the question it settled is reopened. The reasoning is preserved rather than deleted.

**`.planning/SESSION-HANDOFF.md`** — another session's working document, recording the same decision historically.

## Two things this surfaces, needing your decision

1. **The milestones still say 5.0.0.** `5.0.0 — security` and `post-5.0.0` carry the old number. Renaming them is a GitHub-side action, not a code change.
2. **The MAJOR classification and 4.9.0 disagree.** `VERSIONING.md` classes the #322 fix — removal of the documented `wp_sudo_action_replayed` hook — as MAJOR, and that is what drove 5.0.0. At 4.9.0 a documented hook removal ships in a MINOR. That is a real versioning-policy question, flagged in the note rather than resolved here.

Also worth knowing: #381 corrected a planning doc *to* 5.0.0 a few hours ago, so another session may be working from the old number.

## Gates

`verify:metrics`, `verify:sources` (31 citations), `verify:i18n` (POT in sync), `phpcs`, `psalm`, and the 1226-test unit suite all pass.


dknauss:
Heads-up: #389, #390, and #391 are all doing this same rollback, opened within ~6 minutes of each other in response to #392. Comparison and a consolidation recommendation are in [#392](https://github.com/dknauss/Sudo/issues/392#issuecomment-latest) — short version, #389 covers all 15 files (the only one that also fixes `CHANGELOG.md` and `.planning/SESSION-HANDOFF.md`), so it is the one to land. Flagging so three lanes do not keep paying for the same change.

dknauss:
Closing in favour of **#389**, per the comparison in #392.

Not a quality judgement on the change itself — the version-string edits here are correct. It is scope: this PR is deliberately "version string only", which leaves `docs/release-status.md`'s `## Why 5.0.0 and not 4.9.0` justification in place with the number flipped. That is the specific outcome #392 flagged — the premise (that #322 removed the only call site of `wp_sudo_action_replayed`) is false either way, and leaving it standing lets the next release re-derive MAJOR from it.

It also leaves the `pre-5.0.0` string in `CHANGELOG.md` and the `5.0.0` reference in `ActionGatingCompletenessTest.php`, both of which #389 catches.

#389 covers all 15 files and rewrites the rationale. Landing one PR rather than three also avoids the conflict pile-up — three sessions opened this same rollback within six minutes.

dknauss:
## Triage — one fixed here, one deliberately deferred to #393

**Complete the version-reference sweep (P2) — correct, and a real miss.** All three fixed:

- `CHANGELOG.md`'s Unreleased entry told users "pre-5.0.0" sessions need one reauthentication after upgrade. That is the **user-facing** upgrade note, naming a version that will never ship.
- `docs/release-status.md` still read "Stable tag 5.0.0 on main" further down, contradicting the version facts corrected earlier in the same document.
- One test comment.

Worth recording *why*: my survey grep excluded `CHANGELOG.md` as "history". The Unreleased section is not history — it is a current claim about what the next release does — so the exclusion is what hid the user-facing error. Same shape as the other misses on this branch: a scoped search that quietly removed the thing worth finding.

**Handle database versions left at 5.0.0 (P2) — correct, and I am not fixing it here.** Filed as #393.

The behaviour is real: an install stamped `5.0.0` during the window `main` carried it will skip every future `4.9.0` routine, permanently and silently. But `Upgrader::maybe_upgrade()`'s `>=` guard is a **documented contract**, asserted by `UpgraderTest::test_skips_when_version_is_newer()`, and it protects a genuine case — a real downgrade where newer schema is already in place and re-stamping could re-run routines against migrated data.

I prototyped the re-stamp. It is ~8 lines, and it fails that test — which is exactly the point at which it stopped being a version-string change. CLAUDE.md requires a design review before modifying a documented contract, so it belongs in its own change with its own review rather than riding along here.

Scope of the exposure, for the record: no routine exists above `4.0.0` and no `v5.0.0` tag was ever cut, so this is latent rather than live and reachable only on developer installs. #393 lays out four options and recommends the one that adds a Site Health signal without touching the migration path — the defect is that the state is *invisible*, not that it is wrong.

Issue #308, opened 2026-07-26
core-gate: flat 15-min freshness lets a trivial action authorize the most consequential ones


dknauss:
**Severity: HIGH** (Fable). spec §4.2/§11-Q3 recommend flat freshness + optional scope; the PoC calls `wp_has_recent_auth()` with **no scope**. One benign challenge (own-email change, routine plugin update) opens a single 15-min (+~2-min grace) window that satisfies **every** catalog entry — `core/create-application-password` (a durable credential outliving the window), `grant_super_admin`, `core/set-registration-policy`, `install_package()`, `wp_edit_theme_plugin_file()`. This contradicts §5.3's 'gate `grant_super_admin()` **unconditionally**' and the headline that 'possession of an ordinary admin session becomes **insufficient** to introduce executable code' (it is sufficient for ~17 min after any benign gated save), and maximizes the value of the two accepted residuals (XSS free-ride / cookie-stolen-while-open). **Fix (already in the design):** use the `consequence.scope` field; require code-execution and credential-issuance classes to hold their own scope in v1.

---
_Source: Fable+Opus two-model design review of the core-gate proposal/spec (2026-07-26). Doc-anchored claims verified by grep; WP-internal claims to be re-checked against live core before landing._

dknauss:
**Panel → resolved by Shape B.** Per-action step-up removes the reusable recent-auth window for consequential actions, so flat-freshness cross-authorization is **eliminated by construction**. If a window ever returns (Shape A was drafted and rejected), the signed scope must be a compound versioned key v1|{site_id}|{class} to avoid spending the multisite dimension §4.2 promised (U-11). See PR #337.

dknauss:
Resolved in #367 (Unit 1b) — **by construction**.

§4.2 drops the reusable window entirely, so there is no flat freshness left to cross-authorise with: a proof is minted for one action digest, is single-use, and is discarded on redemption. No benign challenge can satisfy a code-execution or credential-issuance action, because no challenge satisfies any action other than the one it was minted for. consequence.scope is therefore no longer load-bearing for v1, and §11-Q3 (scope-bound vs flat freshness) is closed with it.

Recorded for the future: if a reusable window is ever reintroduced, the signed scope must be a **compound, versioned key** (v1|{site_id}|{class}) so the multisite dimension §4.2 had promised is not spent on consequence-class grouping.

PR #403, opened 2026-07-27
fix(upgrader): forbid routines keyed at or below the 5.0.0 stamp (#393)

### LINKED ISSUE #393: Upgrader: a stored db_version ahead of the runtime strands the install
Raised in review of #391. Deliberately **not** fixed there, because it changes a documented contract and that needs a design review first (CLAUDE.md).

> ### ⚠️ Correction — an earlier revision of this issue published an unsafe remedy
>
> It told operators to run `wp option delete wp_sudo_db_version`, and gave a per-site
> `xargs` loop for multisite. **Do not run either.** Both claims were wrong, and design
> review caught them:
>
> - **Multisite is network-scoped, not per-site.** `Upgrader::get_db_version()` uses
>   `get_site_option()` when `is_multisite()`, so there is **one stamp per network**.
>   `wp option delete` never touches `wp_sitemeta`, so the loop deletes a row the
>   Upgrader does not read, changes nothing, and looks like it worked.
> - **Deleting the stamp replays every routine from `0.0.0`** — including
>   `upgrade_2_0_0`'s unconditional `remove_role( 'site_manager' )` (destroys any role
>   with that slug, whoever created it) and `upgrade_3_3_0`'s governance backfill, which
>   grants four capabilities to **every administrator** when it finds zero holders. On a
>   site that deliberately revoked governance and relies on scoped
>   `WP_SUDO_RECOVERY_MODE`, that is a silent privilege escalation caused by a command
>   this issue recommended.
>
> The remedy section below has been replaced.

**Where else this is written down:** `docs/release-status.md` carries a short operator-facing note describing the artifact and its scope (added in #398). That note is deliberately *descriptive only* — it defers disposition here, and carries no remedy command, because the first one published was unsafe (see the correction above). **This issue is the canonical treatment; the release note is a signpost.** Keep it that way rather than copying the analysis into both — `release-status.md` itself warns about the three-copy trap.

## The behaviour

`Upgrader::maybe_upgrade()` returns immediately when the stored `wp_sudo_db_version` is `>=` the runtime `WP_SUDO_VERSION`:

```php
if ( version_compare( $stored, WP_SUDO_VERSION, '>=' ) ) {
    return;
}
```

An install whose stamp is *ahead* of the runtime never advances again. `tests/Unit/UpgraderTest.php::test_skips_when_version_is_newer()` asserts exactly this — `update_option` is never called — so it is an intended contract, not an oversight.

## Why it surfaced

`main` briefly carried `5.0.0` and has been rolled back to `4.9.0` (#391). Any install that loaded `main` during that window has `wp_sudo_db_version = 5.0.0` persisted, and will skip every routine added for `4.9.0`.

Impact is bounded: no routine exists above `4.0.0`, and no `v5.0.0` tag was ever cut, so this is latent and reaches developer/staging installs only.

## Not the same defect as #404 — do not close either as covering the other

Security filed **#404** for `upgrade_3_3_0()`'s zero-holder backfill. Verified independently against `origin/main`: the routine queries holders of `manage_wp_sudo` with `'number' => 1`, and on `empty( $holders )` adds every `Admin::GOVERNANCE_CAPS` entry to every administrator.

The two issues share an origin (the retracted remedy above) but have **different triggers**:

| | #393 (this issue) | #404 |
|---|---|---|
| Trigger | stamp is **ahead** of runtime | stamp is **absent or `0.0.0`** |
| Effect | future routines skipped | `upgrade_3_3_0()` re-runs and re-grants governance |
| Fixed by | the keying rule below | not fixed by it — the routine is keyed `3.3.0`, so a replay from `0.0.0` runs it whatever the stamp *was* |

So "no operator action required" is true here and **false** for any install whose stamp went missing — a restore that dropped the row, a staging→prod copy, or someone who ran the removed command before it was pulled. Those installs have already replayed.

The deeper point is #404's, not mine: `upgrade_3_3_0()` cannot distinguish *"predates governance caps"* from *"deliberately has none"* — both read as zero holders — and it fires no audit event either way. That is reachable on any fresh activation (`Plugin::activate()` calls `maybe_upgrade()` with `$stored = '0.0.0'`), entirely independent of the stamp bug.

## The actual fix: a keying rule, not a stamp rewrite

Design review found the cheap fix, and the repo has already learned this lesson once — see the `HISTORY` docblock on `upgrade_3_3_0()`, where a routine keyed at `3.1.0` never ran for sites stamped `3.1.1`–`3.1.3` and silently locked admins out. The fix then was to **re-key the routine above every stamp in the wild**, not to rewrite stamps.

The same applies here, with one sharp wrinkle:

> Because the guard is `>=`, a routine keyed at **exactly `5.0.0`** will never run on a poisoned install — **not even after `5.0.0` is genuinely released** (stored `5.0.0` == runtime `5.0.0`, guard returns).

So the rule is: **no future upgrade routine may be keyed at any version `<= 5.0.0`.** Key the next one `5.0.1` or higher. The latent defect then expires on its own once runtime exceeds `5.0.0` — no stamp rewriting, no contract change, no operator command.

Enforceable with one unit test asserting no `UPGRADES` key is `<= 5.0.0`, plus a note in the "HOW TO ADD A NEW UPGRADE" docblock at the top of `class-upgrader.php`, which is where the next person actually looks.

## Secondary surfaces (optional, in priority order)

1. **WP-CLI warning** when the condition holds. This reaches the affected audience — developers who live in the CLI. `class_exists( '\WP_CLI' )` is an established legitimate runtime guard here.
2. **Site Health test** at `recommended`. Note the limit found in review: `Site_Health` is only instantiated under `is_admin()`, and core's scheduled check runs in cron, so a test added there does **not** affect the Site Health score, the dashboard widget, or the status email. It is visible only to someone who opens the page. Useful for the non-CLI operator; not a fix.

`recommended`, not `critical`: `test_gated_action_integrity()` is only `recommended` even when a filter has left dangerous actions **ungated right now**. A latent dev-only condition cannot outrank that, and the same signal fires on any legitimate downgrade.

## Rejected

**A corrective migration** (re-stamping down when `stored > runtime`) reverses the documented contract and breaks `test_skips_when_version_is_newer()`. The guard protects a real case: a genuine downgrade where newer schema is already in place, and re-stamping could let a later upgrade re-run routines against migrated data. The keying rule makes it unnecessary.

## If you must repair a specific install

Set the stamp to the running version — do **not** delete it. Deleting replays all seven routines; setting has zero re-run surface.

- Single-site: the option is `wp_sudo_db_version` in `wp_options`.
- Multisite: the same key in **network** options (`wp_sitemeta`), one per network.

Verify the exact WP-CLI invocation against a live install before running it anywhere shared — this issue previously published an unverified command and it was wrong.

dknauss:
Closes #393.

## The fix is a keying rule, not a stamp rewrite

`main` briefly carried `5.0.0` before rolling back to `4.9.0` (#391), so dev/staging installs that loaded it have `wp_sudo_db_version` stamped `5.0.0`. Because `maybe_upgrade()`'s guard is `>=`:

- a routine keyed **below** 5.0.0 never runs there while the runtime is ≤ 5.0.0;
- a routine keyed at **exactly** 5.0.0 never runs **at all** — not even once 5.0.0 genuinely ships, because `stored == runtime` still returns early.

This repo already solved this class of bug once. `upgrade_3_3_0()`'s `HISTORY` docblock records a routine keyed at `3.1.0` that never ran for sites stamped `3.1.1`–`3.1.3` and silently locked admins out; the fix was **re-keying above every stamp in the wild**, not rewriting stamps.

So: a unit test asserting no `UPGRADES` key falls in the dead band, plus a note in the **"HOW TO ADD A NEW UPGRADE"** docblock — where the next person actually looks, rather than in an issue they will never open.

Mutation-verified: keys of `5.0.0` **and** `4.9.0` both red the test with the reason in the message.

`maybe_upgrade()` is untouched. The corrective-migration option stays rejected — it reverses the contract `test_skips_when_version_is_newer()` asserts, and that guard protects a real case (a genuine downgrade where newer schema is already in place).

## What this replaces — a remedy I published that was unsafe

An earlier revision of #393 told operators to run `wp option delete wp_sudo_db_version`, with a per-site `xargs` loop for multisite. **Both were wrong**, and the issue now carries a retraction:

- **Multisite is network-scoped.** `get_db_version()` uses `get_site_option()` when `is_multisite()` → `wp_sitemeta`. `wp option delete` never touches it. The loop deletes rows the Upgrader does not read, changes nothing, and looks like it worked.
- **Deleting replays everything from `0.0.0`**, re-firing `upgrade_2_0_0`'s unconditional `remove_role( 'site_manager' )` (destroys any role with that slug, whoever created it) and `upgrade_3_3_0`'s backfill, which grants four governance capabilities to **every administrator** when it finds zero holders — silent privilege escalation, in a security plugin, from a command we recommended.

The pre-implementation design review caught both **before** any code was written. That is the CLAUDE.md process doing exactly what it is for, and it is why this PR contains a keying rule instead of the Site Health test I originally proposed.

## PoC README

Two changes, both from other sessions' findings:

**Stop citing a closed issue.** #308 closed as resolved-by-construction when §4.2 moved to per-action step-up. Per the Docs session, the README now states what the slice is *evidence about* — a copied login cookie cannot inherit a proof, and `install_package()` fires after `unpack_package()` — both of which hold under per-action step-up too. It no longer reads as endorsing the TTL model.

**Two traps recorded**, per Coordination:

- *Drive the entry point, not the sink.* The first version of these tests called `install_package()` directly, which proves the check works **when reached** and says nothing about **whether it is reached** — exactly how the unpack ordering went unnoticed until a reviewer read `run()`.
- *`git diff A B` is not a merge preview.* It reports main-only files as deletions. `git merge-tree $(git merge-base A B) A B` answers the real question. This cost a false alarm that `poc/` was about to be deleted by a stale branch.

## Gates

1227 tests / 3619 assertions; metrics in sync (six coupled rows updated); sources 31 citations; i18n POT in sync; phpcs; psalm.

### DIFF
--- a/docs/current-metrics.md
+++ b/docs/current-metrics.md
@@ -9,8 +9,8 @@ Verification environment: local workspace, PHP 8.x
 
 | Metric | Value | Verification |
 |---|---:|---|
-| Unit tests | 1,226 tests | `composer test:unit` |
-| Unit assertions | 3,611 assertions | `composer test:unit` |
+| Unit tests | 1,227 tests | `composer test:unit` |
+| Unit assertions | 3,619 assertions | `composer test:unit` |
 | Integration tests in suite | 238 test methods | `rg -c "function test" tests/Integration/*.php | awk -F: '{sum+=$2} END{print sum}'` |
 | Unit test files | 38 | `ls tests/Unit/*.php | wc -l` |
 | Integration test files | 31 | `ls tests/Integration/*.php | wc -l` |
@@ -19,11 +19,11 @@ Verification environment: local workspace, PHP 8.x
 
 | Metric | Value | Verification |
 |---|---:|---|
-| Production PHP lines (`includes/`, `wp-sudo.php`, `uninstall.php`, `mu-plugin/`, `bridges/`) | 20,656 | `find ./includes ./wp-sudo.php ./uninstall.php ./mu-plugin ./bridges -type f -name "*.php" -print0 | xargs -0 wc -l | tail -1 | awk '{print $1}'` |
-| Tests PHP lines (`tests/`) | 42,821 | `find ./tests -type f -name "*.php" -print0 | xargs -0 wc -l | tail -1 | awk '{print $1}'` |
-| Production + tests PHP lines | 63,477 | sum of the two rows above |
-| Test-to-production ratio | 2.07:1 | `42821 / 20656` |
-| Total repo PHP lines (excluding `vendor/`, `vendor_test/`, `.tmp/`, `.git/`, `.claude/`) | 65752 | `find . -type f -name "*.php" ! -path "*/vendor/*" ! -path "*/vendor_test/*" ! -path "*/.tmp/*" ! -path "*/.git/*" ! -path "*/.claude/*" -print0 | xargs -0 wc -l | tail -1 | awk '{print $1}'` |
+| Production PHP lines (`includes/`, `wp-sudo.php`, `uninstall.php`, `mu-plugin/`, `bridges/`) | 20,667 | `find ./includes ./wp-sudo.php ./uninstall.php ./mu-plugin ./bridges -type f -name "*.php" -print0 | xargs -0 wc -l | tail -1 | awk '{print $1}'` |
+| Tests PHP lines (`tests/`) | 42,858 | `find ./tests -type f -name "*.php" -print0 | xargs -0 wc -l | tail -1 | awk '{print $1}'` |
+| Production + tests PHP lines | 63,525 | sum of the two rows above |
+| Test-to-production ratio | 2.07:1 | `42858 / 20667` |
+| Total repo PHP lines (excluding `vendor/`, `vendor_test/`, `.tmp/`, `.git/`, `.claude/`) | 65,800 | `find . -type f -name "*.php" ! -path "*/vendor/*" ! -path "*/vendor_test/*" ! -path "*/.tmp/*" ! -path "*/.git/*" ! -path "*/.claude/*" -print0 | xargs -0 wc -l | tail -1 | awk '{print $1}'` |
 
 ## Footprint & Performance
 
--- a/includes/class-upgrader.php
+++ b/includes/class-upgrader.php
@@ -29,6 +29,17 @@ if ( ! defined( 'ABSPATH' ) ) {
  *    that introduces the change.
  * 2. Add the version string to the UPGRADES array, mapping it to the method name.
  * 3. The method will run exactly once for sites upgrading from an older version.
+ *
+ * ⚠️ 4. **Key it ABOVE 5.0.0.** `main` briefly carried version 5.0.0 before rolling
+ *    back to 4.9.0 (#391), so developer and staging installs that loaded it have
+ *    `wp_sudo_db_version` stamped 5.0.0. Because maybe_upgrade()'s guard is `>=`:
+ *      - a routine keyed below 5.0.0 never runs there while the runtime is <= 5.0.0;
+ *      - a routine keyed at *exactly* 5.0.0 never runs at all, not even once 5.0.0
+ *        genuinely ships, because stored == runtime still returns early.
+ *    Enforced by UpgraderTest::test_no_upgrade_routine_is_keyed_at_or_below_the_poisoned_stamp().
+ *    This is the same fix as upgrade_3_3_0()'s HISTORY note — re-key above every
+ *    stamp in the wild rather than rewriting stamps, which would reverse the
+ *    downgrade contract test_skips_when_version_is_newer() asserts. See #393.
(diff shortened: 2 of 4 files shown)


dknauss:
Cross-reference so neither issue closes as covering the other: this PR closes #393 (stamp **ahead** of runtime → future routines skipped) and **does not** address #404 (stamp **absent/`0.0.0`** → `upgrade_3_3_0()` replays and re-grants governance caps to every administrator).

Verified independently: the routine is keyed `3.3.0`, so a replay from `0.0.0` runs it regardless of what the stamp previously was. The keying rule here governs which routines run *given* a stamp; it has no effect when there is no stamp. #393's body now carries the same split.

PR #407, opened 2026-07-27
docs(developer-reference): add the replay-eligibility contract 4.9.0 promises (#322)

dknauss:
Found by the **Development** session reviewing #398, routed direct per the coordination protocol.

## The gap
#398's release notes tell integrators whose custom rule stopped replaying to *"see the replay-eligibility contract in `docs/developer-reference.md`"*.

```
git grep -c "replay-eligibility\|replay eligibility" -- docs/developer-reference.md   # 0
```

**It doesn't exist.** The disclosure is good and points at nothing, so the only way to learn the rule is to read `Request_Stash` source.

## The contract, from the code
Replay requires the confirmation to have named the **whole effect**:

1. Every effect-bearing allowlisted field is one of the **19** `TARGET_PARAMS`
2. Every other field is in `NON_EFFECT_FIELDS`
3. The target value is not truncated
4. The target is a non-empty array (#397)

I verified every element against `origin/main` before writing — including the parameter lists. **My first extraction truncated the list; Development's was right** (19 params, ending `checked`, `new_role`).

## The worked example is now the breaking case
`developer-reference.md` L42–75 allowlists `item_id`, which is in neither list — so as written that rule reauthenticates correctly and **does not auto-replay** on 4.9.0. I annotated it rather than quietly fixing it: it's the shape people copy, and silently correcting it would leave every existing copy of that rule unexplained.

## The remedies are ordered, and the third is a refusal
Express the effect through a recognised parameter (best — it also makes the confirmation name the thing being changed); or accept reauth-then-resubmit via `post_mode => 'none'`; **or do not** work around it by allowlisting fewer fields so the payload looks complete — that makes the confirmation name less than the request does, which is the exact failure the check exists to prevent.

Docs-only. `verify-sources` 31 citations live; docs-lint clean.

### DIFF
--- a/docs/developer-reference.md
+++ b/docs/developer-reference.md
@@ -45,6 +45,12 @@ is present, the user can still reauthenticate but the POST body is not replayed
 automatically. Use `post_mode => 'none'` for uploads, file-editor saves, or
 other requests that cannot be safely reconstructed from POST fields alone.
 
+Declaring a `stash` policy is **necessary but no longer sufficient** for replay —
+see [Replay eligibility](#replay-eligibility) below. The example that follows
+allowlists `item_id`, which is *not* a recognised target parameter, so as written
+it reauthenticates correctly but **does not auto-replay** on 4.9.0. It is kept as
+the worked example of the breaking case because it is the shape people copy.
+
 ```php
 add_filter( 'wp_sudo_gated_actions', function ( array $rules ): array {
     $rules[] = array(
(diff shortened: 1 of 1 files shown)


dknauss:
Closing — this duplicates work that already exists. **#397 (`fix/322-empty-target-guard`) already contains the replay-eligibility contract**, written in the same commit as the guard. Verified directly rather than on report:

```
git show origin/fix/322-empty-target-guard:docs/developer-reference.md | grep -in "replay eligibility"
# 48: **Replay eligibility requires the confirmation to name the whole effect (#322).**
```

So #398's pointer is forward-referencing a section landing one PR earlier in the agreed order (#397 → #398), not pointing at nothing. Merging this would have produced a **third** copy of one contract.

**How I got here, since it's the exact failure I wrote a rule about a few hours ago.** I verified the absence on `main` and the #398 branch — the two places the message that reached me pointed at — and did not check other open branches. My verification was scoped by the claim instead of by the question. "Read the repo, not the mailbox" (#406) is not satisfied by running a command; it is satisfied by choosing the command's scope independently of what you were told to look for.

**Two things survive and are handed to #397's owner rather than re-landed here** (posted separately):
1. The worked example at L42–75 still allowlists `item_id`, which is in neither `TARGET_PARAMS` nor `NON_EFFECT_FIELDS`, so under 4.9.0 that documented rule no longer auto-replays. #397 does not annotate it. People copy worked examples.
2. #397's recognised-set list appears incomplete: it names 12 params plus "the critical option keys", which omits **`checked`** and **`new_role`**. The constant has 19 entries.

PR #440, opened 2026-07-27
fix(sources): five rows anchored on a file path, which a file never contains (#377)

dknauss:
#377 made the enclosing-symbol column machine-checkable — every backticked token in it must appear at or before the cited line. Five rows ended in a backticked **file path**, and a file does not contain its own path, so those anchors could never resolve.

They have failed on `main` since #377 merged (650a451). `verify-sources.yml` is `schedule` + `workflow_dispatch` and deliberately not a PR gate, so it was invisible on #377 and would have landed as a red `main` on tonight's nightly, attributable to no PR.

The snippets are all still present — every row still says what it claims. Only the anchors fail.

| Row | New anchor |
|---|---|
| `GB-USER-NEW-SELFPOST` | `createuser` — the form's `name`/`id`, on the cited line |
| `GB-ADMIN-POST-BLANK` | `empty( $action )` — the authenticated branch at L61 (the nopriv branch has the same test at L37, so the anchor resolves there first; said so in the column) |
| `GB-NETWORK-EDIT-REDIRECT` | `$action = $_GET['action'] ?? '';` at L13 |
| `GB-USER-EDIT-DIES` | `if ( ! defined( 'IS_PROFILE_PAGE' ) ) {` at L21 |
| `GB-UPDATE-NEEDS-ACTION` | `wp_enqueue_script( 'wp-a11y' );` at L20, immediately above |

Each verified by reading the file, not grepping it — `wordpress-develop` trunk, read 2026-07-27. For every row I checked that the snippet is still unique after whitespace squeezing and that the new anchor's first occurrence is at or before the cited line.

**One row also moves its citation.** `GB-NETWORK-EDIT-REDIRECT` cited the assignment at L13; it now cites the guard itself at L15, `if ( empty( $action ) ) {` — unique in the file, and what the claim is actually about. The assignment becomes its anchor. The claim keeps its note that `wp_redirect( network_admin_url() );` cannot be cited directly, occurring at both L16 and L40.

**Why anchors rather than the exemption list.** Two of these five have no named enclosing symbol at all — a top-level statement, a `<form>` — which is the category `UNANCHORED_LEGACY_IDS` documents, and extending that list was the other honest answer. But the list exempts a row from the anchor check entirely, whereas every one of these files does supply something real to anchor on: the construct the cited line tests, or the statement immediately above it. That is the shape #377 itself used for `GB-PIN-STAR`. The prose says which of the two each row uses, so nobody has to infer it.

`bin/verify-sources.sh` on this branch: **52 citations checked, all snippets present**, no failures — down from 5.

"How to add a row" now states that a file path is not an anchor, so the next row does not repeat the mistake.

Scope note: this touches `docs/upstream-sources.md` only — no `bin/verify-sources.sh`, no `tests/verify-sources/run.sh` — deliberately disjoint from `fix/377-anchor-red`, which is live in another worktree and addresses qualified-anchor containment. Coordinated with the Development session before starting.

Refs #377.


dknauss:
Superseded by the Development session's `fix/anchor-rows-nightly-red` (`dcbb4c3`), which was pushed while this was open. Closing in its favour — its version is better on every row I can check, and two of my anchors were wrong in a way worth recording rather than quietly dropping.

**Where mine was weaker.** I anchored on tokens that *precede* the cited line; theirs *enclose* it, which is what the column is for:

- `GB-USER-NEW-SELFPOST` — mine was `createuser`, whose first occurrence is L189, nowhere near the L519 form. Theirs is `if ( current_user_can( 'create_users' ) ) {` at L513, which genuinely opens the block containing it.
- `GB-ADMIN-POST-BLANK` — mine was `empty( $action )`, first occurrence L37. That is the **unauthenticated** branch; the row is about the authenticated one at L61. I noticed the double occurrence and wrote it into the prose instead of treating it as the defect it was.
- `GB-USER-EDIT-DIES` — mine was `if ( ! defined( 'IS_PROFILE_PAGE' ) ) {` at L21, which precedes the chain but does not open it. Theirs anchors the chain itself at L25.

**And their exemption call is right where I was wrong.** For `network/edit.php:13` and `update.php:22` I reached for adjacency — `wp_enqueue_script( 'wp-a11y' );` at L20 — on the GB-PIN-STAR precedent. That precedent does not transfer: GB-PIN-STAR's className is on the same element as its cited line, whereas `wp_enqueue_script` is an unrelated statement that happens to sit above. It would satisfy the checker while proving nothing about containment, which is the vacuous-needle failure this whole lane exists to close. Their `TOP_LEVEL_STATEMENT_IDS` — a separate list from `UNANCHORED_LEGACY_IDS`, because "predates the anchor requirement" would be a false statement about a row written this week — is the honest answer.

**Two things from here worth salvaging, offered rather than pressed:**

1. **A note in "How to add a row"**, so the next row does not repeat this. Nothing in the file currently says a path is not an anchor:

   > The **backticked** tokens in the symbol column are what the checker looks for in the file, so backtick only things that appear *in* it. A file path is not one of them: five rows once ended `` … in `wp-admin/update.php` ``, and since a file never contains its own path, no anchor could resolve.

2. **`GB-NETWORK-EDIT-REDIRECT` could cite a better line**, independently of the anchor question. It cites the assignment at L13 while its claim is about the redirect guard; `if ( empty( $action ) ) {` at L15 is unique in the file and is what the claim is actually about. Under the exemption it still needs one — nothing encloses L15 either — so this is only about which line the row points at.

Diff kept in this closed PR for reference; branch deleted.

Issue #434, opened 2026-07-27
Guard test: assert HANDLER_ENDPOINTS and $verified_screens are disjoint


dknauss:
Follow-up from [#433](https://github.com/dknauss/Sudo/pull/433) (#429). Not a shipping defect — recording it so the gap is closed deliberately rather than rediscovered.

## The gap

In `ChallengeTest::test_every_builtin_post_rule_lands_somewhere_usable()`, the handler predicate is checked **before** `$verified_screens` and `continue`s on a match:

```php
if ($this->isHandlerEndpoint($url)) {
    continue;
}

if (in_array($url, $verified_screens, true)) {
    continue;
}
```

So a URL present in *both* lists passes via the handler branch and its `$verified_screens` entry is never consulted. Adding a real screen to `Challenge::HANDLER_ENDPOINTS` silently demotes it, and the guard stays green.

Demonstrated during review: adding `'themes.php'` to `HANDLER_ENDPOINTS` leaves the full suite green (1,247 tests) while every refused theme delete/update POST reverts to the dashboard. The same holds for `options-general.php`, `profile.php`, `plugin-editor.php`, `theme-editor.php`, `update-core.php`, `authorize-application.php`, site `edit.php`, and network `settings.php`. (`plugins.php`, `users.php` and `user-new.php` happen to be pinned by the individual landing tests.)

## Why it was not a blocker

It fails in the **safe** direction: the user lands on the neutral dashboard with the notice intact, which is the pre-#429 behaviour. No dead end, no lost notice, no security consequence. That is the opposite of the fail-open direction the guard is scoped to in its docblock, and it requires someone to *add* an entry to a hand-curated const that carries a per-entry justification.

## Fix

One assertion — the two sets must be disjoint:

```php
foreach ($verified_screens as $url) {
    $this->assertFalse(
        $this->isHandlerEndpoint($url),
        "$url is listed as a verified screen but classified as a handler."
    );
}
```

## Also worth doing while in there

`array_diff` on bare filenames means a `pagenow` used by *both* a site rule and a network rule would only be tested at the site base. No such value exists today, so this is pre-emptive.

PR #441, opened 2026-07-27
fix(sources): five rows anchored on a file path that never resolves

dknauss:
`main` has been red on `composer verify:sources` since #377 merged as 650a451 — five rows fail the anchor check. This fixes them: **rc 1 with 5 failures → rc 0 with 0**.

### Why they broke

#377 made the Enclosing symbol column a **conjunction**: every backticked span must resolve at or before the cited line. These five were authored under the old OR semantics and carry a backticked **file path** as one span. A path does not appear inside its own file:

| row | path occurrences in its own file |
|---|---|
| GB-ADMIN-POST-BLANK, GB-NETWORK-EDIT-REDIRECT, GB-USER-EDIT-DIES | 0 |
| GB-USER-NEW-SELFPOST | only L521, *after* the L519 snippet |
| GB-UPDATE-NEEDS-ACTION | only a docblock past the cited L22 |

Under OR another span carried each row. Under AND the path fails. Snippets are all still present — anchors only.

`verify-sources.yml` is `schedule` + `workflow_dispatch`, deliberately not a PR gate, so this surfaces as a **red nightly attributable to no PR** — the standing red that workflow's own header warns turns into a check people stop reading.

### Three rows name their real enclosing construct

- `GB-USER-NEW-SELFPOST` → `` `if ( current_user_can( 'create_users' ) ) {` `` (L513 encloses the cited L519 form)
- `GB-ADMIN-POST-BLANK` → `` `if ( empty( $action ) ) {` `` (L61 encloses L67)
- `GB-USER-EDIT-DIES` → the `if`/`elseif` chain opened at L25. The cell names the **chain**, because the cited L27 is a sibling `elseif`, not nested inside L25's block — a bare anchor there would assert containment that does not hold.

### Two cite top-level statements, where no enclosing symbol exists

`network/edit.php:13` has only comments and one `require_once` above it. `update.php:22` is itself the conditional wrapping the file body (L22–L372 of 372).

Any token nameable there — `IFRAME_REQUEST` at L9, `wp_enqueue_script` at L20 — sits above the cited line but **encloses nothing**, so it would satisfy the checker while proving nothing about the claim. That is the vacuous-needle failure this lane has spent its time removing, so these get an exemption instead.

Deliberately **not** `UNANCHORED_LEGACY_IDS`: that list warns the row "predates the anchor requirement", which is false for rows written this week, and would put a false statement in the gate's own output. New `TOP_LEVEL_STATEMENT_IDS` with its own message — "exempt by kind, not by age" — and a test asserting the message does *not* claim they are legacy.

The exemption is **anchor-only**: snippet presence and drift still run (test 22b pins it).

### Verification

- `composer verify:sources` — rc 0, 0 failures, both exempt rows WARN and the three anchored rows are silent (passing the check, not skipping it)
- `bash tests/verify-sources/run.sh` — 117 passed, 0 failed
- The new exemption branch is mutation-verified: removing it reds test 22
- All line numbers read from live `wordpress-develop` trunk on 2026-07-27

### Known limitation, unchanged

Two of the three new anchors also occur earlier in their files, so the check proves "such a construct exists at or before the cited line", not that *this* one encloses it. That is the containment limit the checker already discloses and these rows cannot escape without a parser.


dknauss:
Disposition on the open P2 (top-level exemption keyed on row ID): **accepted as a real defect, deferred to #442.**

It genuinely fails open — `is_top_level_statement` returns true from the row ID alone, so a row later repointed at a different source keeps its exemption without anything re-establishing that the cited line is still top-level. That is a cached judgement with no way to notice the world moved, which is the same shape as the `HANDLER_ENDPOINTS` gap in #434 and three of the defects found during #433 review.

Deferring on timing, not merit. `main` has been red on `composer verify:sources` since #377 and the tag gate requires it green; this PR is the narrow repair for that. Binding the exemption to a validated marker is a second change to the same subsystem inside the same window, and the standing call on this subsystem going into the release is to stop expanding it — the immediate defect is small relative to the churn already around it.

#442 records the finding, the reasoning, and that it wants fixing together with #434 since they are the same class.

Resolving so the merge is not blocked. Reopen the thread if you disagree with the deferral rather than the finding.

dknauss:
Triage of the three open threads — one applied, two deferred with reasoning.

**P1 (false absolute) — applied in `a9a232c`.** Correct, and I verified it rather than taking it: `wp-admin/user-new.php` contains its own path **twice** and `update.php` once, both in docblocks. That is the falsifying command `AGENTS.md` requires any *never* to survive, and this one did not — in the document that governs scope-word discipline.

The absolute was also load-bearing for the wrong reason. A path fails as an anchor because finding it establishes nothing about what encloses the cited line, not because it is absent. **Presence is not containment** — the same reason a nearby statement is not an anchor, which the paragraph already said two sentences later.

**P2 (top-level workflow) — partly applied, rest deferred to #442.** The gap is real: following the documented workflow for a third top-level citation produces a row the checker rejects, because `is_top_level_statement()` knows only two hardcoded IDs. `a9a232c` documents the allowlist step so the workflow is at least followable. Making the exemption genuinely kind-based is #442, deferred on timing — `main` is red on `verify:sources` until this merges and that is a tag gate.

**P2 (qualified anchors) — deferred to #442.** Plausible and worth acting on: if `expand_anchor_spans()` resolves `Configurator::extract_assoc()` without requiring it literally, then "backtick only things that appear in the file" understates what the checker accepts and could lead someone to replace a qualified identity with a weaker declaration-shaped token. I have not verified the expansion semantics against the script, so I am not writing guidance about them under a freeze — asserting how a checker behaves without reading it is the failure this registry exists to prevent. Recorded in #442 rather than guessed at.

Resolving both P2s. Reopen if you disagree with the deferral rather than the finding; #442 carries them.

Issue #452, opened 2026-07-27
capture_target() has no rule context, so stored-option suppression applies to every rule


dknauss:
Deferred from PR #444 (raised by Codex review; confirmed by two independent reviewers). Real, **fail-closed**, not a security defect.

## The defect

`Request_Stash::capture_target()` has the signature:

```php
private function capture_target( bool &$complete ): array
```

and is called as `$this->capture_target( $target_complete )`. No rule ID and no rule array is passed in — so it has no idea which rule matched.

PR #444 added `target_value_echoes_stored_option()`, which suppresses a target field when its submitted value equals the stored option, keyed on `Request_Stash::CRITICAL_OPTION_SOURCES` (`siteurl`, `home`, `admin_email`, `new_admin_email`, `default_role`, `users_can_register`). Because `capture_target()` cannot tell rules apart, that suppression runs for **every matched rule**, not only `options.critical`.

## Consequence

A custom rule whose effect field is literally named one of those six — a plugin with its own `home` or `siteurl` field carrying unrelated semantics — that submits a value equal to the corresponding stored WordPress option will have that field silently dropped from the target.

Traced consequence: the suppressed key is absent from `$target`; `target_describes_payload()` uses `array_key_exists()` per payload key, so it returns `false`; `Challenge::may_replay_bound_stash()` then refuses the replay.

So the action is **discarded rather than replayed** — the user reauthenticates successfully and lands on the fail-closed path, then re-issues the action manually.

## Why this is a wart, not a vulnerability

It fails **closed**. Nothing executes that the confirmation did not name; the cost is a lost replay and a manual re-submission. The dangerous direction would be a suppression that hid a change while still replaying it, and that cannot happen here — the same predicate that hides the field also refuses the replay.

That is why it was deferred out of #444 rather than fixed there: the obvious fix threads the matched rule through `capture_target()`, a signature change on a security-relevant path, and #444 was the last change before the v4.9.0 tag.

## Fix sketch

Pass the matched rule into `capture_target()` and apply the equality suppression only for `options.critical` — or, more generally, only for rules that declare their fields as option-shaped.

Worth deciding deliberately rather than mechanically, because the narrower behaviour is not obviously right in all cases: a custom rule that genuinely gates a WordPress option would *want* the suppression. A rule-declared opt-in (`'compare_stored_options' => true`, say) may be better than hard-coding the rule ID, and it keeps the registry the single place that knows what a rule's fields mean.

## When picking this up

[...]

Issue #431, opened 2026-07-27
The Target: line shows raw internals ('users: 2, new_role: author') on the Users-list role change — the control it exists to provide


dknauss:
Found by hand-testing 4.9.0. **This is #322 v2's headline control failing in a common flow**, in exactly the way the code's own comment says it must not.

## What a user sees

Users list → change a user's role → the challenge appears:

```
To continue: Change user role — please enter your password.
Target: users: 2, new_role: author
```

`users: 2` is a bare database ID. `new_role: author` is a raw form field name. A human being asked to authorise something cannot tell from this **which account** is affected.

The tester's suggested wording is the right shape: *"Demote **someone** to Author"*.

## Why this is not cosmetic

`Challenge::describe_stash_target()` already knows this. Its own comment:

> A bare numeric id ("user_id=5") tells an admin nothing about WHO they are authorizing — **and this line is the only control left in the same-origin lure case**, so resolve it to something a human can actually judge.

So the design explicitly identifies bare IDs as inadequate *because* the Target line is the last line of defence when an attacker lures a user into a gated request. In that scenario the user is shown `users: 2` and has no way to judge it.

The whole argument for #322 v2 is that a coarse label is a blank cheque and naming the concrete target makes the password submit an informed decision. Here the target is named in a form only the database understands.

## Root cause

The resolution is keyed to a single field name:

```php
if ( 'user_id' === $key && current_user_can( 'list_users' ) ) {
    $user = get_userdata( (int) $value );
    if ( $user && ! empty( $user->user_login ) ) {
        $value = $user->user_login . ' (#' . (int) $value . ')';
```

Only `user_id` is resolved. The Users-list role change posts **`users[]`**, so `$key` is `users` and the branch never fires. `user.promote`'s own stash allowlist confirms the field names:

```php
'stash' => self::stash_allowlist( array( 'users', 'new_role', 'changeit' ) ),
```

So the resolution and the allowlist disagree about what a user identifier is called. The lookup was written for `users.php?action=promote&user_id=N` (the row-action path) and never extended to the bulk/dropdown path.

## What a fix needs

1. **Resolve `users` as well as `user_id`**, including the multi-value case — the field is an array, so several accounts can be affected by one confirmation, and naming only the first would be worse than naming none.
2. **Render field names as human labels.** `new_role: author` should read as a role name, not a form key.
[...]

Issue #429, opened 2026-07-27
After reauth, POST actions land on the dashboard and the notice tells you to re-submit a form that isn't there


dknauss:
Found by hand-testing the 4.9.0 challenge flow. **Not a security defect** — the no-replay behaviour is correct and deliberate. It is a usability failure in how that correct behaviour is delivered, and in one common flow the user's work is simply lost with no path back.

## What a user experiences

1. Go to **Users → Add New**, fill in the form (username, email, password), submit.
2. The gate intercepts and shows the challenge. Fine.
3. Enter the password, reauthenticate successfully.
4. **You land on the Dashboard.** Not the Add New User screen.
5. A notice says: *"Reauthentication complete. For your security, password and secret fields were not replayed. **Re-enter them** to finish the change."*

Re-enter them **where?** There is no form on the Dashboard, and everything typed in step 1 is gone. The instruction describes an action the landing page cannot support.

Reported verbatim by the tester: *"the user will assume it's the password reauth form they just completed"* and *"creating a new user completely fails and ends up outside the new user creation screen."*

## Mechanism

`Challenge`, in the post-reauth redirect construction:

```php
$neutral_url = is_network_admin() ? network_admin_url() : admin_url();
$target      = $neutral_url;

if ( 'GET' === strtoupper( (string) ( $stash['method'] ?? 'GET' ) ) && ! empty( $stash['url'] ) ) {
    // …return to the originating screen with the query stripped…
}
```

**GET** actions return to the originating screen. **POST** actions fall through to `admin_url()` — the dashboard. The code comment states it as intended: *"POST actions land on the dashboard (their re-submit needs the form re-filled regardless)."*

That reasoning is half right. The form does need re-filling — but the user cannot re-fill a form they are no longer on.

Create-user reaches this path via `may_replay_bound_stash()` refusing with `redacted_fields`: the submitted password is deliberately dropped at stash time, so origin-bound replay (#322 v2) correctly declines. **The security logic is working.** The delivery is what fails.

## Two defects, separable

**1. The landing page (the functional one).** A POST should return to its originating *screen* — `user-new.php` for create-user — the same way GET already does. The query still has to be stripped (the action + nonce is the effect, and returning to the action URL is the confused-deputy hole #322 v1 closed), but the screen itself is safe and is where the instruction points.

**2. The copy (both variants).** Both notices assume the reader is looking at the form:

- `"Review the form and submit it again to finish the change."`
[...]

Issue #413, opened 2026-07-27
stash_no_replay() is ignored for GET and bodyless POST (network.theme_disable is replayed despite declaring no-replay)


dknauss:
## Summary

`Request_Stash::build_stashed_post_params()` returns **before** applying the rule's stash policy when the request method is not `POST`, or when `$_POST` is empty. A rule that explicitly declares `stash_no_replay()` therefore has its refusal ignored on those paths, and the action is auto-replayed after reauthentication.

This is the rule's own declared policy being skipped — distinct from #412, where a rule's target fails to name the effect. #412 is "the confirmation is wrong"; this is "the rule's refusal is not consulted."

## Mechanism

`includes/class-request-stash.php`, in `build_stashed_post_params()`:

```php
if ( 'POST' !== $method || empty( $_POST ) ) {
    return array();
}

$stash_policy = $this->get_stash_policy( $matched_rule );
$post_mode    = $stash_policy['post_mode'];

if ( 'none' === $post_mode ) {
    $post_replay_blocked      = true;
    $post_replay_block_reason = self::REPLAY_BLOCKED_NO_REPLAY;
    return array();
}
```

The early return precedes the policy check, so `$post_replay_blocked` stays `false`. `Challenge::may_replay_bound_stash()` consults that flag, finds no block, and `build_replay_response_data()` proceeds to replay — for a GET, by redirecting to the stashed URL, which executes the action.

## Scope — one live rule, verified

Of the nine builtin rules declaring `stash_no_replay()`, eight set `'method' => 'POST'` in their admin matcher and are not reachable by GET:

`plugin.upload`, `theme.upload`, `user.promote_profile`, `user.change_password`, `user.change_email`, `auth.app_password`, `editor.plugin`, `editor.theme`

One does not:

- **`network.theme_disable`** — `'method' => 'ANY'`, `'pagenow' => 'themes.php'`, `'actions' => array( 'disable', 'disable-selected' )`, callback `is_network_admin()`, `'stash' => self::stash_no_replay()`

So a network-admin GET to `themes.php?action=disable&theme=…` is gated, stashed, and then **replayed after reauth despite the rule declaring no-replay**.

[...]

PR #455, opened 2026-07-27
fix(challenge): decouple the unchanged-value map from wp_sudo_critical_options

dknauss:
Follow-up to #444. **Not a new defect** — these are review-thread fixes that missed the merge.

## Why this is a separate PR

#444's auto-merge fired the moment its review threads were resolved, which was before I pushed the fixes those threads asked for. GitHub then deleted the branch, so the push landed on a recreated branch attached to nothing.

#451 has since landed the two documentation corrections, and **its versions are better than mine** — the `developer-reference.md` paragraph keeps `new_admin_email` as the historical example and cites `GB-ADMIN-EMAIL-FIELD`, which mine dropped. So this PR carries only what is genuinely still missing from `main`.

### 1. The comparison map is still derived from `wp_sudo_critical_options` (the real one)

Raised by Codex on #444. Deriving the unchanged-value map from the **filtered** critical list reinstates #431 on exactly the sites that customised their protection set:

- a site narrows `wp_sudo_critical_options` to just `new_admin_email`;
- an ordinary Settings → General save still posts unchanged `siteurl`, `home`, `default_role`, `users_can_register`;
- those keys remain in `TARGET_PARAMS` but fall out of the filtered map, so `target_value_echoes_stored_option()` returns false and names them as changes again.

That is #431's exact bug, under a filter. Default configuration is unaffected, which is why it survived review of the original PR.

The instinct was wrong, not just the code: I derived from the filtered list for "parity" with `critical_option_rest_keys()`. But **whether an option is *gated* and whether a displayed value is *true* are different questions**, and the same form posts the unchanged fields either way. `critical_option_comparison_map()` becomes a private const `CRITICAL_OPTION_SOURCES`, which also drops the `Action_Registry` dependency from this path.

Regression test added — filter narrowed to `new_admin_email`, asserting only the changed option is named. Mutation-checked: restoring the filter-derived map turns it red.

### 2. A false comment on the `options.critical` rule

The comment claimed the rule lands the user "back on the settings screen". It does not. `options.php` is in `Challenge::HANDLER_ENDPOINTS`, so `is_handler_endpoint()` returns true and the landing is the **neutral dashboard** — deliberately, because returning a refused Settings-API POST to `options.php` renders the raw All Settings dump (`GB-OPTIONS-ALLSETTINGS`).

The corrected comment also records that this landing is **pre-existing**, not caused by `stash_no_replay()`: `new_admin_email` sat in the stashed payload and never in the target, so `target_describes_payload()` was already false and replay already refused.

### 3. The deferred limitation is undocumented in code

`capture_target()` has no rule context, so the stored-option suppression runs for **every** matched rule. A custom rule whose effect field is literally named `home`/`siteurl` and whose submitted value equals that stored option has it suppressed, after which `target_describes_payload()` refuses the replay.

Fails **closed** — nothing executes unnamed; the user re-submits manually. A functional wart for third-party rules, not a security defect, which is why it was deferred rather than fixed under the freeze. Now filed as #452, with a `KNOWN LIMITATION` block that names the issue and instructs its own removal so it cannot outlive the fix.

### 4. A stale symbol in the citation registry

The `GB-ADMIN-EMAIL-FIELD` row still named `critical_option_comparison_map()`, deleted in (1). Caught by review, not by tooling — `verify-sources` checks the **upstream** snippet and enclosing symbol, while the "Claim it supports" column is free prose about **our** code, which nothing validates.

## Gates

| Gate | Result |
|---|---|
| `composer test:unit` | 1,266 tests / 3,768 assertions |
| `composer lint` | 24/24 |
| `composer analyse` | PHPStan L6 + Psalm clean |
| `composer verify:metrics` | in sync |
| `composer verify:sources` | 57 citations, all present |
| `bash bin/make-pot.sh` | regenerates to no diff |

## Note for whoever sequences the tag

`main` currently carries the filter-derived map, so a site using `wp_sudo_critical_options` sees the #431 false Target line. Worth landing before the tag if the window is still open; if not, it is a narrow, non-default, fail-safe display bug and can follow.


dknauss:
**Before trusting `main`'s green: it has no test for the affected path.**

I verified `main` (`7e9812a`) in a detached worktree with a full `composer install` — 1,265 tests / 3,766 assertions, lint, PHPStan L6 + Psalm, `verify:metrics`, `verify:sources` and an idempotent `.pot` all pass. That green is real, and it says nothing about the defect this PR fixes. The filtered-map regression only surfaces when `wp_sudo_critical_options` is narrowed, and the test that exercises that path (`test_narrowed_critical_options_filter_does_not_restore_false_target`) exists **only on this branch**. `main` has no coverage of the filtered configuration at all, so it is green *and* carrying the bug. Nothing looked, so nothing objected — which is not the same as having checked. Whoever picks this up after the tag should read the suite's scope before reading its result: a green signal whose scope is narrower than it appears is the pattern that ran through this entire release, and this is a clean instance of it.

dknauss:
**Disposition on the three open threads — one applied, two superseded.**

**"Update the metrics verification date" — applied** in `602d550`. `docs/current-metrics.md` still read `Last verified: 2026-07-26` while this branch moves the unit counts, so the file asserted a verification that did not happen on that date. `verify:metrics` checks the numbers and not the date, which is why nothing caught it — the same perishable-figure class #457 was about.

**"Refuse GET replays with suppressed effect fields" and "Preserve comparisons for filter-added critical options" — superseded, not dismissed.**

Both are refinements to *replay eligibility*: which stashes may be auto-executed after reauthentication under a narrowed `wp_sudo_critical_options`. The release has been re-scoped to a single invariant:

> After reauthentication, WP Sudo never automatically executes a previously intercepted server-side request.

Global — every method, every rule, every surface. Under it there is no eligibility question to get right, because nothing is eligible. A suppressed effect field cannot cause a wrong GET replay when no GET is replayed.

That is the whole reason for the re-scope: #397, #412, #413, #429 and #444 were each individually fixable, and collectively they showed that *replay eligibility* is the wrong question to keep answering. These two threads are two more correct answers to it.

**What survives on this branch is the part that is not about replay** — the comparison map no longer being filter-derived, so a narrowed filter stops displaying unchanged options as changes. That is a display defect, live on `main` today for non-default configurations, and it fails safe by over-statement.

**If the invariant does not land, reopen both.** They are accurate against the current mechanism and I am recording that explicitly rather than closing them as wrong.

Resolving so the merge is not blocked.

Issue #412, opened 2026-07-27
Replay eligibility: a GET rule whose effect rides an unrecognised param can be replayed via a recognised decoy (#322 residue)


dknauss:
Follow-up from #397. The empty-target guard landing there closes one half of the informed-confirmation control; this is the other half. Filed with the design analysis while it is fresh, because two candidate fixes were tried and refuted and that work should not be repeated.

## The defect

For a custom **GET** rule whose real effect rides a parameter not in `Request_Stash::TARGET_PARAMS`, an attacker appends a *recognised decoy*:

```
?action=custom&item_id=evil&id=1
```

1. `capture_target()` reads `$_GET` for `TARGET_PARAMS`, records `id => 1`, silently ignores `item_id`.
2. `$complete` stays true — nothing was truncated.
3. `target_describes_payload( ['id'=>'1'], [] )` returns true after **zero iterations**, because `build_stashed_post_params()` returns `array()` for any non-POST, so a GET stash always has an empty payload.
4. #397's non-empty-target guard passes — the target *is* a non-empty array.
5. Replay proceeds against the stashed URL, which still carries `item_id=evil`.

The challenge said `Target: id 1`. What executed was `item_id=evil`.

## Exposure is narrow, and this is not a regression

- **Unreachable on built-ins.** Every GET-reachable replayable built-in carries its effect in a recognised param — `plugin`, `stylesheet`, `theme`, `users`/`new_role`, `id`. The one exception, `tools.export`, has *no* recognised param and is therefore already refused by the empty-target guard. The bypass requires the effect itself to ride an unrecognised param, which no built-in does.
- **Worse behaviour shipped in 4.8.0.** At the `v4.8.0` tag there was no target naming, no binding proof, and unconditional `'replay' => true`. This is a narrow residue of a hole that was total.

So the exposed population is custom GET rules whose effect field is not a recognised target.

## Two candidate fixes, both refuted — do not retry without reading this

[...]

Issue #435, opened 2026-07-27
Fail-closed landing uses is_network_admin(), which is false on the admin-ajax path


dknauss:
Follow-up from [#433](https://github.com/dknauss/Sudo/pull/433) (#429). Pre-existing, severable, and deliberately left out of the release fix.

## What

In `Challenge::build_replay_response_data()`:

```php
$neutral_url = is_network_admin() ? network_admin_url() : admin_url();
```

Both `replay_stash()` call sites are inside `Challenge::handle_ajax_auth()` and `Challenge::handle_ajax_2fa()`, registered as `wp_ajax_*`. So this runs under `wp-admin/admin-ajax.php`, which never calls `set_current_screen()` and does not define `WP_NETWORK_ADMIN`. `is_network_admin()` reads `$GLOBALS['current_screen']`, then `WP_NETWORK_ADMIN`, then returns `false` — so it is **always false here**, even for a genuine network action.

Consequence: a refused network-admin action lands on the **site** dashboard rather than the network dashboard.

## Why it was not fixed in #433

The notice still renders. `render_blocked_replay_notice()` is hooked to both `admin_notices` and `network_admin_notices`, and the site dashboard fires `admin_notices` — so the user gets the explanation, just on the wrong dashboard. A wrong-dashboard wart, not a lost explanation, and unrelated to what #429 reported.

The #433 handler classification originally *also* keyed on `is_network_admin()`, which made it dead code; that was fixed there by deciding from the URL path instead. Nothing in the classification path consults request context any more, so this is now the only remaining consumer and it is severable.

## Fix

Derive the same way the handler check does — from the stashed URL's path — or thread the surface through from the caller. Worth a look at whether any other `is_network_admin()` / `is_multisite()` call reachable from the ajax handlers has the same problem; two instances of this exact fault were found during #433 review, plus one in a test.

## Related

- #434 — guard-test disjointness follow-up from the same review.

dknauss:
**The premise in this title is now known to be false** — retracted by [#488](https://github.com/dknauss/Sudo/pull/488) and recorded as entry #72 in `docs/llm-lies-log.md`.

[...]

PR #459, opened 2026-07-27
fix(322): never auto-execute an intercepted request after reauthentication

### LINKED ISSUE #322: Security (P1): stash auto-resume is a confused deputy — a planted stash + victim reauth executes the attacker's transaction

dknauss:
Closes #322. Refs #412, #413.

## The invariant

> **After reauthentication, WP Sudo never automatically executes a previously intercepted server-side request.**

Every method, every rule, every surface. `build_replay_response_data()` has no branch that can return a replay instruction — the guarantee is verifiable by reading one method.

**7 files, +239 / −346.** Net negative. This replaces an abandoned 36-file rewrite (~2,246 / 2,115) that tried to remove every trace of replay, rewrite all documentation and reshape the test suite in one security release. That branch is preserved as reference and not merged.

## What changed

The stash is **always** consumed, the binding cookie **always** cleared, `wp_sudo_replay_refused` **always** fires, and the user lands on the existing safe soft landing holding the session they just earned.

The dead machinery is left in place for removal in 4.10 — with one forced exception: `may_replay_bound_stash()` is deleted, because PHPStan L6 rejects an unused private method and its own guidance forbids suppressing the error.

## Why removal rather than repair

An interim design authorised replay per-request behind a `__Host-` binding cookie plus a confirmation naming the target. Sound in outline, repeatedly wrong in detail:

- guards keyed on values that could not hold where they ran (#429)
- a target line naming an unchanged field (#431)
- a route where replaying a partial Settings-API body made core write `null` over every option the body omitted

Each was individually fixable. Together they showed that *which stashes may replay* is the wrong question to keep answering. **Removal fails closed.**

## One addition beyond the brief — flagging for rejection if you disagree

The "session already confirmed" page auto-navigated to `$_GET['return_url']` on `DOMContentLoaded` **with no password step**. A user already holding a sudo session who opened a crafted challenge URL was navigated to an address of the requester's choosing — the same confused deputy, reached without touching the stash.

Against the attacker this release is about — one holding a cloned cookie, and therefore able to mint a valid nonce (`GB-NONCE-TOKEN`) — that is live.

Left in place, the headline claim would be **false by a route its wording does not cover**, which is worse than not making the claim. The **Continue** link already on the page is the explicit second action.

## Version: MINOR

`wp_sudo_action_replayed` is **retained but dormant** — hook, documentation and every subscriber intact, so nothing an integrator wrote errors, but nothing fires it.

MINOR rather than MAJOR because the signal has a **successor at the same lifecycle point**: `wp_sudo_replay_refused` fires for every consumed stash carrying a `reason`. The consumer's work is a rename, not a loss. *(An event made dark with no successor would be MAJOR — `VERSIONING.md` does not currently name retained-but-unreachable at all.)*

Refusal vocabulary collapses from ten eligibility reasons to `replay_disabled` and `no_credential_this_request`, the latter being the lure footprint and the only value worth alerting on.

## Tests

One data-provider case per required proof — GET, POST, bodyless POST, active-session resume, Application Password authorization — each asserting no `replay` / `post_data` / `replaying` key and no redirect to the action URL.

**The binding cookie is present in every case**, deliberately: under the old mechanism that was the thing that authorised a replay, so the invariant is proven with the old authorisation *satisfied*.

Nineteen eligibility-branch tests were removed along with the contract they encoded, rather than weakened — a half-asserted old guarantee is worse than none.

**Mutation-tested:** restoring POST replay and restoring the GET redirect each turn the suite red.

## Verification

`test:unit` 1267 / 3780 · PHPCS · PHPStan L6 + Psalm · `verify:metrics` · `verify:sources` (57 citations) · `verify:i18n` — all green.

**Integration and E2E not run locally** (no MySQL, no wp-env). CI is their first execution.

Audit-hook count drops 22 → 21, which is the dormant hook confirming it has no producer.


dknauss:
## Disposition of all 16 open threads

Triaged against the current head. Most predate the code they describe — this PR was rewritten three times (classifier → blunt rule → test corrections), so a thread filed against an earlier push often describes something that no longer exists.

### Applied in this PR

| Thread | Disposition |
|---|---|
| Sanitize the active-session cancel target | **Fixed, then superseded.** `return_url` is no longer read anywhere in `Challenge`; both cancel URLs are the server-chosen neutral page. |
| Narrow the no-replay invariant to exclude editor re-dispatch | **Fixed.** Wording narrowed to *server-stashed* requests everywhere, with the editor's in-tab re-dispatch explained rather than hidden. |
| Avoid classifying every active-session resume as a lure | **Fixed.** The server cannot distinguish a multi-tab resume from a lure landing on a session-holder, so the docs now say exactly that: correlate, do not alert. |
| Retire the replay-availability Site Health test | **Fixed.** `test_replay_binding()`, its registration, its three unit tests and its strings are removed — it was false in both branches. |
| Finish the replay-removal prose sweep / Remove remaining replay-era documentation | **Fixed** across `class-challenge.php`, `developer-reference.md`, `release-status.md`, `CHANGELOG.md`, `readme.txt` and `two-factor-integration.md`. |
| Replace the now-false automatic-progress message | **Fixed.** The `replayingAction` string is removed; the client announces only the honest message. |
| Treat disabling the documented hook as a breaking release | **Reasoning replaced.** The original justification was a non-sequitur and you were right to call it. `VERSIONING.md` now carries a *Security-forced inertness* clause, and this release is its worked example. Maintainer ruling: MINOR. |
| Preserve the public API return flow | **Accepted as a documented behaviour change.** `wp_sudo_require()`'s `return_url` is inert — retained so callers do not fatal, consumed by nothing, disclosed at the symbol's own entry, in the changelog and in release-status. |
| Remove the mutually exclusive URL assertion | **Fixed** in `0d88a09`. That was my bug: a count-based replacement hit STACK-01 instead of STACK-05. STACK-01's original assertion is restored. |
| Update the tracked E2E count | **Fixed.** 107 → **112 tests in 20 files**, with the LAND-01..05 rationale recorded. |
| Correct the test-to-production ratio calculation | **Fixed.** The expression divided numbers that were not the tracked inputs. |

### Deliberately not fixed here

| Thread | Disposition |
|---|---|
| Give GET actions an applicable retry instruction | **Real, live, and filed as #463.** Every stash now takes the refusal path, so a link-driven GET action is told to "review the form" it never had — the most common path after this change. Not fixed here only because it is a production string, and this PR's production code is byte-identical to the SHA two independent reviewers cleared. **First out after the tag.** |
| Preserve valid admin.php plugin screens | **Moot for the cancel target** — it is unconditionally neutral now, so no classification decides it. **Correct by design for the consumed-stash landing**: the effect-bearing query is stripped *before* the handler check, so what remains is bare `admin.php`, which renders nothing. The retained `page` parameter the thread describes does not survive stripping. |
| Reject bare-path cancel targets | **Moot.** There is no requester-supplied cancel target. |
| Keep network-handler landings in network admin | **Pre-existing, filed as #435.** `is_network_admin()` is false under `admin-ajax.php`, so network actions land on the site dashboard. Unrelated to replay; the notice still renders there via `admin_notices`. |

Resolving all sixteen. Reopen any where you disagree with the disposition rather than the finding.

PR #464, opened 2026-07-27
chore(release): point the Playground blueprint at v4.9.0

dknauss:
Post-tag step for **v4.9.0** (`82d4b23`).

The public *Try latest release* Playground badge loads `blueprint.json` from `main`, so this bump has to happen **after** the tag exists — doing it before serves a ZIP that 404s.

Verified the target resolves before committing:

```
curl -sIL https://github.com/dknauss/Sudo/archive/refs/tags/v4.9.0.zip
→ 200
```

One line. No plugin code.


dknauss:
Applied — you were right, and it is the same class of defect this release kept surfacing.

The blueprint bump moves the public Playground demo to `v4.9.0`, and `docs/release-status.md` — the canonical release-state document — still said the latest tag was `4.8.0` and that `4.9.0` was unreleased. Merging this alone would have pointed the demo at a release the document of record denied existed.

Corrected in this PR rather than deferred: latest tag is now `v4.9.0` on `82d4b23`, `4.8.0` moves to *previous*, and the `4.9.0` payload is recorded — including the two disclosed behaviour changes (`wp_sudo_action_replayed` dormant with its successor named; `wp_sudo_require()`'s `return_url` inert) and the *Security-forced inertness* clause that classifies them as MINOR.

For the record on scope: this PR publishes nothing. The `v4.9.0` tag and its GitHub Release already exist — pushing the tag triggered `release.yml`. This only repoints the demo blueprint at the tag that is already public, which is the documented post-tag step.

Issue #436, opened 2026-07-27
Post-reauth: the work is still lost, the notice still misdescribes it, and one form's changes are named for only one rule


dknauss:
Hand-testing the four flows after #429 merged (`eb6fe3f`). **The landing fix works** — flows now arrive where Coordination predicted. What the fix exposes is that landing on the right screen was necessary and not sufficient.

Filed as one issue because these are three faces of the same journey; split if that is easier to schedule.

## 1. Returning to the form does not return the work

**Users → Add New** now lands back on the Add New User screen, exactly as intended. But every field the user typed is gone, and the notice reads:

> Reauthentication complete. For your security, password and secret fields were not replayed. **Re-enter them** to finish the change.

"Re-enter **them**" implies the secrets specifically. In practice username, email, first/last name, website — everything — must be retyped. The user is returned to a correct but empty form and told a partial truth about why.

Tester's verdict: *"this will never be an acceptable UX in core or anywhere."*

Non-secret allowlisted fields are already stashed (`stash_allowlist()` carries them). Repopulating them on the landing screen would leave only the password to retype, which is what the notice already claims. That is a materially smaller ask than replay — no action is executed, a form is pre-filled.

## 2. The notice text is wrong on nearly every landing

The same string appears wherever a POST is refused:

> Reauthentication complete. For your security, this request was not replayed automatically. **Review the form and submit it again** to finish the change.

Observed landings for that string:

| Flow | Lands on | Is there a form? |
|---|---|---|
| Edit another user | Dashboard | **No** |
| Settings → General / Settings → Sudo | Dashboard | **No** |
| Plugin activate | Plugins list | No form — a **link** to re-click |
| Users → Add New | Add New User | Yes, but empty |

Three of four landings have no form at all. The tester's original point stands and is not addressed by #429: having just completed a password form, *"the form"* most naturally reads as the one they just submitted.

The copy needs to vary by landing: re-click the action / re-enter the form / your changes were not saved. It cannot be one string when the destinations differ this much.

## 3. A form that triggers several rules is described by only one

Changing **role, password, and email together** on `user-edit.php` produced:

> To continue: **Change user role** — please enter your password.

Three gated rules matched (`user.promote_profile`, `user.change_password`, `user.change_email`); the challenge names one.

Cause: `$action_label = $stash['label'] ?? $stash['rule_id']` — a stash carries a single `rule_id`.

[...]

PR #468, opened 2026-07-27
fix: three 4.9.0 follow-ups — inert return_url signal, notice copy, guard test

### LINKED ISSUE #434: Guard test: assert HANDLER_ENDPOINTS and $verified_screens are disjoint

### LINKED ISSUE #461: wp_sudo_require(): _doing_it_wrong() when the now-inert return_url is passed
Follow-up from [#459](https://github.com/dknauss/Sudo/pull/459). Suggested by an independent reviewer; deliberately **not** done in the security release because it is production code and would have invalidated both reviewers' clearances.

## The gap

As of 4.9.0, `wp_sudo_require()`'s `return_url` argument is **inert** — still accepted, still emitted into the challenge URL, consumed by nothing. Automatic navigation to a requester-supplied destination after a successful challenge executes under the sudo authority just minted, so it had to go.

The behaviour change is documented in `CHANGELOG.md`, `readme.txt`, `docs/developer-reference.md` (at the symbol's own entry) and `docs/release-status.md`, and the classification now has a rule behind it (*Security-forced inertness* in `VERSIONING.md`).

**But an integrator who passes `return_url` today gets no runtime signal.** Their users land on the dashboard instead of the expected page, and nothing anywhere says why. They have to find the changelog and connect it themselves.

## Proposal

Call `_doing_it_wrong()` when `return_url` is passed non-empty:

- Turns a silent behavioural regression into a loud one, in the place the integrator is actually looking (their debug log).
- Costs nothing when the argument is absent, which is the overwhelmingly common case.
- Standard WordPress idiom for exactly this situation.

Needs a test asserting it fires only when the argument is non-empty, and does **not** fire for the default.

## Why this is worth doing rather than leaving

The reviewer's framing is the right one: silence is the failure mode hardest to notice, and this is the one remaining place in the 4.9.0 change where an integrator gets silence rather than a signal. Everything else — the dormant hook, the removed replay — either has a successor firing at the same lifecycle point or produces immediately visible behaviour.

### LINKED ISSUE #463: Post-reauth notice tells GET actions to "review the form" — there is no form
Found by Codex review on [#459](https://github.com/dknauss/Sudo/pull/459). **Live user-facing defect, introduced by the no-replay change**, filed rather than fixed because #459 is a security release under a scope freeze and this is a production string change.

## The defect

`Challenge::render_blocked_replay_notice()` says:

> Reauthentication complete. For your security, this request was not replayed automatically. **Review the form and submit it again** to finish the change.

Before 4.9.0 only POSTs reached that notice, so "the form" was accurate. **Now every stash takes the refusal path**, so a link-driven GET action — plugin activation, theme activation, anything reached by clicking a link — lands on its list screen carrying `wp_sudo_blocked_replay=1` and is told to review a form it never had.

This is the **most common** path after the change, not an edge case. The single piece of recovery guidance the user gets is inapplicable to the action they just took.

## Fix

Either wording that covers both shapes:

> …this action was not repeated automatically. Perform it again to finish the change — you will not be asked to reauthenticate.

or select the notice by request method, the way the redacted/blocked variants are already selected. The second is more precise and there is precedent for it in the same class.

Worth adding: the current copy also does not say the sudo session is already active, which is the reassuring half. The retry passes straight through, and the user has no way to know that.

## Why it was not fixed in #459

The PR is frozen to the invariant. Its production code is byte-identical to the SHA two independent reviewers cleared, and changing a shipped string — however safe — would break that property and force another adversarial round for no security gain. The defect is cosmetic in the strict sense: nothing behaves wrongly, the guidance is just wrong for one request shape.

**It should be first out after the tag.** It is the sentence most users will see.

dknauss:
Closes #461, closes #463, closes #434.

All three were deferred out of #459 because it was a security release under a scope freeze and each touches production code or a user-facing string.

## #461 — `wp_sudo_require()`'s inert `return_url` now says so

`return_url` has been inert since 4.9.0 (#322). An integrator passing it got no runtime signal: their users land on the dashboard and nothing anywhere says why. Now raises `_deprecated_argument( 'wp_sudo_require', '4.9.0', … )`.

Three decisions worth review:

- **`_deprecated_argument()`, not the `_doing_it_wrong()` the issue proposed.** The caller made no mistake — they passed a still-documented argument that stopped doing anything. Note four `AdminTest` assertions pin `_doing_it_wrong` as `->never()` for a *different* case; that precedent does not transfer, and [class-admin.php:3430](includes/class-admin.php:3430) says why ("the option name was never a callable"). Here the subject *is* a callable and the misuse is its argument.
- **`$version` is `'4.9.0'`.** Core documents `_deprecated_argument()`'s `$version` as "the version of WordPress that deprecated the argument used" — when it went inert. That is a *different* convention from `_doing_it_wrong()`'s "version where the message was added". A design review applied the latter to the former; corrected against core source.
- **Handle is the literal `'wp_sudo_require'`.** `WP_UnitTestCase::deprecated_function_run()` keys `$caught_deprecated` by `$function_name`, so the documented `setExpectedDeprecated( 'wp_sudo_require' )` remedy only works with the literal — `__METHOD__` would have broken it.

**This will redden integrator test suites that pass `return_url`.** That is the intent, and the suppression string is documented at the symbol's own entry per `VERSIONING.md`'s inertness clause.

Fires before every early return, so it does not go quiet for integrators whose users are usually already elevated.

## #463 — the post-reauth notice no longer names a form

It told everyone to "review the form and submit it again" — true before 4.9.0, when only POSTs reached it. Now four landings reach it and none presents *the form for the action being repeated*.

Took the issue's option (a), and **rejected its preferred (b)** (select by request method). (b) does not work: `is_handler_endpoint()` diverts a POST to `options.php` or `user-edit.php` to the dashboard too, so method does not decide what the user finds. (a) also avoids changing the emitted value of a `public const` that 8 unit assertions and 8 E2E URL matchers depend on.

The claim is deliberately narrow. These screens mostly *do* have forms — `plugins.php` and `users.php` render bulk-action forms, the dashboard renders Quick Draft — which is why naming one misdirects rather than merely being vague. An earlier draft of this change said "three of the four have no form"; that was false, core was read, and the sentence narrowed.

`notice-warning` retained deliberately, reasoning in the docblock.

## #434 — guard test for an invariant nothing enforced

`HANDLER_ENDPOINTS` and the verified-screens list make contradictory claims about the same URL, but the walk checked the handler predicate first and `continue`d, so overlap was never detected — adding a real screen to `HANDLER_ENDPOINTS` silently demoted every refused POST on it while the suite stayed green.

**Verified, not assumed:** injecting `themes.php` into `HANDLER_ENDPOINTS` makes the new guard fail and leaves the pre-existing walk passing. Run twice, before and after the test-file edits.

Asserted through the production predicate on full URLs rather than by intersecting filenames, because `is_handler_endpoint()` carries a context clause for network `edit.php`. Reordering the two checks would not fix it — that is the mirror-image bug. `$verifiedScreens` is now a shared method so the walk and the guard cannot diverge.

## Deliberately not changed

`docs/security-test-results-4.8.0.md:43` keeps the old notice string. It records what was *observed* during 4.8.0 live testing, so editing it would falsify an observation rather than update a description.

## Known follow-up, not in this PR

The sibling redacted notice at [class-challenge.php:170](includes/class-challenge.php:170) ("Re-enter them to finish the change") has the same problem: a POST to `user-edit.php` is in `HANDLER_ENDPOINTS` and lands on the dashboard. Same class, in the notice #429 was specifically about. Worth its own issue.

## Validation

`composer test:unit` 1280 tests / 3822 assertions · `composer lint` 24/24 · `composer analyse` PHPStan L6 + Psalm clean · `composer verify:i18n` (POT regenerated with `bin/make-pot.sh`, not hand-patched) · `composer verify:metrics` (tests, assertions and four line-count rows updated, plus a stale `44676 / 21012` ratio formula) · `composer verify:sources` · `git diff --check`.


dknauss:
Closing in favour of refiling. The work is still wanted — **#461, #463 and #434 stay open** — but this branch is no longer the cheapest path to it.

## Why

It branched at `8303509`; `main` has since gained **11 commits**, and a rebase conflicts in seven files:

```
wp-sudo.php                    version bumped to 4.9.1
docs/current-metrics.md        line counts resynced three times
languages/wp-sudo.pot          regenerated
includes/class-challenge.php   heavily edited by #479 and #488
docs/upstream-sources.md       a row added and then removed
docs/developer-reference.md    replay-language sweep
tests/Unit/ChallengeTest.php   docblock corrected in #488
```

Two of those — `current-metrics.md` and `wp-sudo.pot` — are **generated**. They must be regenerated, not hand-merged, and hand-resolving them is exactly how a false count or a stale line reference enters a tree. Reconstructing this on top of current `main` is less work than rebasing it, and carries less risk.

## The better reason

This PR fixes **#463** (`render_blocked_replay_notice()`) on its own. But **#469** (`render_redacted_replay_notice()`) and the second face of **#436** are the *same surface* — the two notice methods sit adjacent in `Challenge`, and both say something false about what happened to the user's work. Landing this separately would collide with whatever fixes the other two, and would leave the user reading one corrected notice and one uncorrected one depending on which path they hit.

Those three want to be **one change**, not three.

## What replaces it

- **#461** (`wp_sudo_require()` inert `return_url` → `_deprecated_argument`) — small and independent; refile on its own.
- **#434** (guard test asserting `HANDLER_ENDPOINTS` and `$verified_screens` are disjoint) — test-only; refile on its own.
- **#463** — folds into the post-reauth notices change with #436 and #469.

The design decisions recorded in this PR's description are worth carrying forward; nothing here is being discarded as wrong, only as stale relative to `main`.

PR #470, opened 2026-07-27
research: reconstruct server-held action approval candidate

dknauss:
## Purpose

Resolve Phase 27 as a falsifiable browser-security decision before any production implementation.

## Result

The tranche reconstructs and exercises a server-held action-approval candidate across Node and real WordPress transports. The accepted boundary is intentionally narrower than an XSS-proof claim: authorization remains server-held and effect-bound, while active same-origin script and full cookie-state cloning remain explicit non-goals or concessions. The decision record stops before Phase 28 production work.

## Included evidence

- Browser handoff and adversarial fixtures with explicit mutation guards.
- WordPress/PHP adapter, real upgrader path, browser transport, concurrency, and failure-path coverage.
- Reproducible Node and WordPress runners plus the Phase 27 CI workflow.
- Guard manifests and a shared guard-id cross-check, including the #494 quote-style drift fix.
- Decision, candidate-boundary, reconstruction, and exit-handoff records.
- Phase 27 multisite evidence is explicitly deferred under #490; it is not represented as complete here.

## Validation at current head

- Worktree-local Composer install and classmap verified.
- `composer test`: 1307 tests, 3904 assertions.
- PHPCS passed.
- PHPStan level 6 and Psalm passed.
- `npm run test:research:phase27:mutations`: baseline passed and all 15 current mutations killed at their named assertions.
- GitHub: Phase 27 evidence audit, Node baseline/mutations, WordPress browser transport/integration/concurrency, PHPUnit matrix, Plugin Check, CodeQL, Psalm, documentation checks, and E2E shards passed after the rebase.

## Merge posture

Draft. The evidence is reproducible and green, but this remains a review-sized research tranche rather than production implementation.


dknauss:
## Independent adversarial review — four agents, pinned to `d3693b7`

Four Claude subagents reviewed this PR independently, each with a single lens, each instructed to **reproduce the evidence and find a counterexample** rather than summarize, and each required to name what it attempted and ruled out. All four verified they were at `d3693b769d9f2a07c71845d0065bddd028818f9c` and left the tree clean.

All four were given the same adversarial frame: **the fixture is simultaneously the system under test and the source of evidence**, so a guard that holds only because the mock implements it proves nothing about WordPress.

Environment integrity, checked independently by each: `node_modules/` and `vendor/` in the worktree are **real directories, not symlinks** into another checkout.

**Headline:** the mutation table is honest and fully reproducible — 18/18 documented mutations run today and fail exactly where the record says, with 0 rows that are merely asserted. The defect is **coverage**, not fabrication. Two un-killed guard clauses yield demonstrated post-approval substitution while the suite stays green.

---

## 1. Mutation validity

**Reproducibility — clean.** Baseline 10/10 pass (4.9s). All **18 documented mutations executed** (16 table rows + 2 negative-control mutations described in prose); **18/18 killed**, each at the assertion the record names. A `diff` of the `PHASE27_DISABLE_*` names present in the fixture against those in the record is **identical** — no confabulated switches, no undocumented ones. The 16 selected-candidate mutants fail at 16 **distinct** assertion lines, so no single assertion is doing all the killing.

Counts requested: **18 confirmed-and-reproduced, 0 "the document asserts this."**

### U-1 — UN-KILLED GUARD with demonstrated bypass (upload path)

The guard is:

```js
uploadIntent.digest !== '' && ( binding !== uploadIntent.preparedBinding || preflight.digest !== uploadIntent.digest )
```

The killing test uses Browser B, which trips **both** disjuncts at once. Delete only `preflight.digest !== uploadIntent.digest` and **all 10 tests still pass**. On a single unchanged browser binding:

| step | pristine | digest clause removed |
|---|---|---|
| preflight trusted bytes | 200 | 200 |
| approve (correct password) | 200 | 200 |
| **re-preflight attacker bytes, post-approval** | **409** | **200** |
| upload attacker bytes | 409 | **204**, `upload-effect` value `407efd44…` |

Approved digest `17f75876…`; executed digest `407efd44…`. The guard that stops Browser A swapping approved bytes after approval exists, works, and **nothing in the suite would notice its removal**. The row labelled "Upload preflight immutability" demonstrates browser binding, not immutability.

### U-2 — same shape, file path, worse

For `/candidate-preflight` the equivalent clauses `preflight.target !== candidateIntent.target || preflight.digest !== candidateIntent.digest` are **unreachable dead code**: the hard-coded allowlist earlier in the same branch pins `action`/`target`/`digest` to single constants, so those disjuncts can never be true when evaluated. Protection against post-approval target substitution comes from fixture scaffolding, not the guard the table credits. Removing the allowlist *and* reducing immutability to its binding clause (each survives alone) yields `candidate-effect` = `attacker-plugin/attacker.php`, status **204**, all 10 tests still passing.

### U-3 — the login-session half of the binding is entirely un-mutated

"Approval binds both" (browser binding + WordPress login-session verifier). Replacing `! authenticated` with `false` in **both** approve endpoints, or in **both** effect endpoints, breaks no test. There is no table row for authentication on any of the four endpoints. Given the headline claim is specifically about what a copied *auth* cookie can and cannot do, the absence of mutation evidence for the auth-cookie requirement itself is a notable hole.

*(Independently corroborated by the upload-approval reviewer, which found the same gap by grepping every browser-context creation in the spec: all set `wp_auth`, so no request in the suite ever reaches an upload endpoint without it.)*

### U-4 — effect-time "exact match" exercised one field at a time

`/candidate-effect` compares intent-id, target and digest; only **target** is exercised. Dropping intent-id *and* digest passes. `/candidate-upload-effect` compares intent-id and digest; only **digest** is exercised — dropping the intent-id clause passes. The contract text ("reconstructs the action descriptor… recomputes any content digest, and atomically consumes only an exact match") is broader than the evidence.

### U-5 — guards with no row and no exercise

Each deletable with 10/10 green: the `binding === ''` rejection in `authorizePreflight`; the `kind` / `^[a-f0-9]{64}$` validation in `/candidate-upload-preflight`; the allowlist in `/candidate-preflight`; the bearer check in `/redeem`. For contrast the password check in `/confirm-and-redeem` **is** killed, so the rejected-candidate fixtures are not uniformly untested.

### D-1 — `/mutate-intent` is a tautology endpoint cited as an observation

That handler unconditionally pushes `{operation:'mutation-blocked'}` and unconditionally returns 409 — no guard, no comparison, no state. The isolated-confirmation test asserts that observation as a result, and the Evidence table's Mutate column for **Isolated top-level** reads "exact server-held intent". That cell is backed by a hard-coded constant. It does not change the decision (that candidate is rejected on other, genuinely observed grounds), but per this project's claim discipline the cell should be marked not-observed, or the endpoint should compare something.

### D-2 / D-3 / D-4 — labels and prose stronger than evidence

- "File preflight immutability" and "Upload preflight immutability" are both killed solely by the browser-binding disjunct. Rename (e.g. "preflight binding") or add the same-binding re-preflight assertion.
- "Browser B cannot overwrite Browser A's **immutable** prepared intent" — immutability is not what is exercised.
- "Approval binds both" — the login-session half has no mutation.
- The summary claim "survive binding, mutation, and one-use guard mutations" **is** fully traceable to killed rows and should stand.
- "The browser receives only status, not a readable bearer" rests on the `HttpOnly` attribute, correct by construction but never asserted — the full-clone test reads the binding via Playwright's context cookie API, which *sees* HttpOnly cookies and so does not test JS-unreadability. One `document.cookie` assertion closes it.

---

## 2. Preflight disclosure

Ran the suite (10/10, 4.9s) and both documented preflight mutants; both killed at the documented points. Also wrote a raw-`fetch` attack harness, since the adversary here is a script holding a cookie, not a browser.

**Ruled out:** timing oracle (median 7.1 / 8.2 / 7.8 ms across unauthorized-403, authorized-unknown-403, authorized-known-200 with a 2 MB body — `authorizePreflight()` does reject before `readBody()`, and the ~1 ms delta only separates which check failed, already visible in the status code); unmetered probing via a missing binding cookie (50 binding-less requests all 403, zero budget consumed, ordering correct); unbounded map growth from rotating `wp_auth` (auth check returns before `preflightAttempts.set()`); descriptor field leakage (no response field is a lookup result).

### Reachable bypass — reauthentication is unthrottled

`preflightAttempts`/`authorizePreflight()` is called from exactly two sites, the two preflight handlers. `/candidate-approve` and `/candidate-upload-approve` have **no attempt budget at all**. Measured, holding only the copied auth cookie:

1. `GET /copied-cookie-candidate` → server mints the attacker their own binding cookie
2. `POST /candidate-preflight` → 200, intent prepared against the attacker's binding
3. **500 sequential wrong-password POSTs to `/candidate-approve` → 500 clean 403s**, no throttle, no lockout, no delay
4. correct password → 200
5. `POST /candidate-effect` → **204**, `used = true`, observation recorded

No `PHASE27_DISABLE_*` toggle covers it, no test asserts it, and the contract specifies rate limiting only for preflight. Production WP Sudo *does* throttle reauth (`Sudo_Session`, 5 attempts → 5-minute lockout), so the shipped plugin is not open — but this document selects a protocol, and the accepted claim is true only under a control the protocol never states and the evidence never models. Either add it to the contract and model it, or narrow the accepted statement to assume a throttled reauthentication step.

*(The cookie-state reviewer independently reached the same observation and classified it a later-phase concern rather than a bypass, reasoning it is reachable by a full-cookie-state clone rather than within trace A's stated capabilities. Both endpoints are named in that finding. The severity split is a genuine scope disagreement, flagged rather than resolved.)*

### Un-killed guard — the probe budget is keyed on a caller-supplied value

`const rateKey = scope + ':' + auth;` where `auth` is the `wp_auth` cookie. The adjacent comment and the doc both call this budget bound "to the authenticated login session, not the attacker-rotatable browser cookie". The binding half is true and well-tested. The session half is not:

- one fixed session token, 6 probes → `[200,200,200,429,429,429]`
- rotating the session token per request, 6 probes → `[429,200,200,200,200,200]`

The budget resets on every distinct session value. The mock is immune only because `authorizePreflight()` accepts exactly one hardcoded string, so nothing *can* rotate. In WordPress the analogue is a per-session token from `WP_Session_Tokens`; a user holds many concurrent sessions and every fresh login mints a new value. **Login-session scope is precisely the scope that re-login resets.**

Note who this splits: the cookie thief holds one stolen session and generally cannot re-login, so the budget binds them. The *authenticated low-privilege insider* — the principal the test itself names `low-privilege-session` — can log in at will and mint unlimited fresh budgets. The design already has the right scope name elsewhere: the approval store keys on "actor, login-session verifier, browser binding".

Honest counterweight: actor scoping is **not** a free fix. A and B share the auth cookie and therefore already share the budget — that sharing is what produces the test's `429`, and it means an attacker can burn the victim's budget and lock the legitimate user out of preparing intents. Actor keying makes that lockout stronger. This is a genuine Phase 28 design tension and should be recorded as an open question, not as a settled property.

Two smaller mechanics: the window is fixed, not sliding (`windowStartedAt` pinned at first attempt), so 3 probes at t=59s and 3 at t=61s pass; and a `429`'d request still increments `attempts` without extending the window.

### Documentation mismatch — "by capability" is string equality

The test is named *preflight withholds descriptors by capability*. The implementation is `auth !== 'copied-login-session'` → 403. No user model, no role, no capability decision — `low-privilege-session` is rejected because it is a *different string*. The fixture cannot distinguish "unauthenticated" from "authenticated but unauthorized".

The doc partially discloses this ("the fixture models the action's capability as a distinct authorized principal … production must substitute each action's real capability decision"), which is right. But the preceding sentence states the result as though two principals were exercised: "anonymous and authenticated low-privilege callers receive the same minimal 403". Nothing authenticates the low-privilege caller. Only the anonymous half of the kill point is real.

Worth stating plainly because it inverts the test's rhetoric: the string that *passes* the gate is literally `copied-login-session` — the credential the threat model concedes to the attacker. The test demonstrates that preflight is **reachable by the cookie thief**; it does not demonstrate withholding from a legitimate lower-privileged user.

### Documentation mismatch — contract response fields vs. fixture

Contract lists opaque intent ID, canonical action ID, server-defined label, canonical target fields, exact digest, expiry. `/candidate-preflight` returns four fields (`digest`, `id`, `label`, `target`); `/candidate-upload-preflight` returns three. No canonical action ID, no expiry. Expiry is explicitly deferred in Limits; the canonical action ID is neither returned nor deferred. IDs are `candidate-file-write-1` / `candidate-upload-1` — self-describing, not opaque. Security impact nil, but contract and evidence should agree.

### The disclosure is in the status code, not any field

Enumerated: `digest` and `target` are echoes of caller input; `id` and `label` are hardcoded server constants. Nothing is a lookup result — so **the withholding property is structurally undemonstrable in this fixture**. There is exactly one descriptor and it is a string literal; there is no store of files, plugins or paths for an existence oracle to interrogate. Flagged as self-referential evidence: the test's name promises a property the mock's shape makes impossible either to exhibit or to falsify.

What the *shape* of the protocol leaks, for a caller past the auth gate:

| probe | response |
|---|---|
| unknown action/target/digest | 403 |
| recognized, no other binding holds it | 200 + full descriptor |
| recognized, another binding has it prepared | 409 |
| budget exhausted | 429 |

The 403/409/200 discrimination is an existence oracle over the `(action, target, digest)` triple. The contract's disclosure control governs the **body** and does nothing here, because the status code answers "is this a valid target?" before any body is returned. In production, where `target` is a real path, that is an enumeration primitive — bounded only by the preflight budget, which is why the keying finding above matters more than it first appears.

`409` additionally tells the caller that another browser session sharing this auth cookie is currently mid-ceremony on this exact target. As a clone block that is a win; the same signal hands the cookie thief the victim's in-flight operation.

---

## 3. Upload approval

**No reachable bypass found.** No way for Browser B, holding only the copied `wp_auth` cookie, to preflight over, approve, or redeem Browser A's intent.

**Byte binding — verified independently, not read.** `/candidate-upload-effect` reads via `readBodyBuffer()`, which pushes raw `Buffer` chunks and never calls `setEncoding` (contrast `readBody()`, `utf8`, used only on JSON endpoints). Digest is `createHash('sha256').update(body)` over content, not metadata. Independently computed SHA-256 of the 8 test bytes `50 4b 03 04 00 ff 80 41` = `17f75876c4a7e94e98d23d54290a11c43d386ea46977a387b6aa249a2e930b01`, exactly the asserted value. A UTF-8 round-trip would instead give 12 bytes `504b030400efbfbdefbfbd41` (0xff, 0x80 → U+FFFD) and digest `3a21911c…` — so **any encoding round-trip, truncation or transcoding would change both length and digest and fail the test**. Preflight validates `/^[a-f0-9]{64}$/`; uppercase, `sha256:`-prefixed and 63-char digests all rejected (probe-confirmed).

The doc's "including `0x00`, `0xff`, and `0x80`, upload once" is literally true and independently reproducible. Note the mutation package differs only in its trailing byte (`0x41`→`0x42`), so the mutation arm does not itself exercise the high bytes — the trusted arm's digest match does.

**One-use atomicity — empirically atomic; could not produce two winners.** Fired 2, 8 and 32 genuinely concurrent redemptions with correct binding, intent and bytes. Every time: **exactly one 204, the rest 409, exactly one observation.** The reason is structural: `await readBodyBuffer()` completes before the guards, and between the `used` check and the `used = true` write there is no `await` — the hash is synchronous, so Node's run-to-completion makes the section indivisible. **That property does not transfer to php-fpm**, where the equivalent read-check-write across separate processes is a genuine TOCTOU. The doc already says so; reviewers should not mistake the green test for design evidence.

**Binding + intent are not separable.** Approval requires binding **and** intent **and** password; redemption requires binding **and** intent **and** digest. Each stage fails independently (probe-confirmed): effect with fresh binding and no approval → 403; correct bytes and binding but wrong/absent intent header → 409; approve without `wp_auth` → 403; effect without `wp_auth` → 403. Re-preflighting a different digest from the approved binding → 409. No constructible ordering satisfies one dimension while violating the other.

**All 7 declared upload mutations killed**, every kill point matching the table exactly (`IMMUTABILITY`→793, `PASSWORD_CHECK`→806, `APPROVAL_BINDING`→819, `APPROVAL_INTENT`→832, `BINDING`→851, `DIGEST_CHECK`→861, `ONE_USE`→884).

### Un-killed guards on the upload path

Four guards present and working (probe-confirmed) whose deletion would leave all 10 tests green:

| Guard | Probe | Why no test kills it |
|---|---|---|
| `! authenticated` on `/candidate-upload-approve` **and** `/candidate-upload-effect` | 403 both | **Every** browser context in the spec sets `wp_auth=copied-login-session`; no request ever reaches an upload endpoint without it |
| `binding === ''` on approve/effect | — | same |
| `preflight.kind !== 'plugin'` | `kind:'theme'` → 403 | spec only ever sends `kind:'plugin'` |
| digest-shape regex | uppercase/prefixed → 403 | spec only sends well-formed 64-hex |

Mutation-coverage gap, not a reachable bypass. The `!authenticated` one matters most: the WordPress login-session dimension is the axis the entire copied-cookie threat model turns on, and it is the one axis the suite never exercises negatively.

Separately, the effect-stage intent-ID correlation has **no independent mutation** — `submittedIntent !== uploadIntent.id` is bundled inside the same `PHASE27_DISABLE_UPLOAD_DIGEST_CHECK` condition as the digest comparison, and no test sends a wrong intent header to `/candidate-upload-effect` (the wrong-intent test targets *approve*). One table row covers two distinct checks. Non-exploitable with a single intent ID; load-bearing the moment multiple intents exist — i.e. Phase 29.

### Transport shape is not among the stated limits

The fixture accepts only a raw `application/octet-stream` body where the body *is* the file. Real WordPress plugin upload (`update.php?action=upload-plugin`) is `multipart/form-data` via `$_FILES`. Confirmed the fixture rejects a multipart-wrapped copy of identical bytes (409). Production must hash the **decoded part** (`$_FILES['pluginzip']['tmp_name']`) and the client must hash the same pre-encoding bytes — a canonicalization step the fixture never exercises. **This is the main place where "the digest binds exact bytes" is trivially true in the mock and non-trivial in WordPress.** One Limits bullet covers it.

---

## 4. Cookie-state boundaries

Full suite 10/10 (5.5s). **All 16 documented mutations run individually — every one kills its named test.** `composer verify:sources` passes (58 citations); the new `GB-WEBAUTHN-CODE-INJECTION` row resolves against live w3c/webauthn source. **No reachable bypass within trace A's stated capabilities.**

**Ruled out: "the fixture assigns different bindings by construction."** It does not. Both `/copied-cookie-candidate` and `/copied-cookie-upload-candidate` mint **only when the cookie is absent** and otherwise echo back what was presented. Test 8's Browser B and test 9's clone navigate the *same page through the same code path* with the *same* `wp_auth`, differing in exactly one input — whether `phase27_binding` was carried. 403 vs 204. A real differential, not a stipulation.

**Ruled out: predicting or obtaining A's binding.** A fresh mint is a 43-char base64url string = `randomBytes(32)`, 256 bits of CSPRNG, taking **no input** from `wp_auth`, the intent, the request, or anything the attacker holds.

**Ruled out: the same-origin-HTML rejection being asserted rather than demonstrated.** The noopener-strip test observes a live cross-window `handle.document` credential read after the fixture patches `window.open` ahead of the application handler; the counterfeit test observes `history.pushState()` faking the path with `parent-unloaded` absent. Genuine browser-behavior observations.

### D1 — the binding is never attested, and the contract is weaker than what this plugin already ships

The fixture adopts **any** client-supplied binding verbatim (`Set-Cookie: phase27_binding=attacker-chosen-binding-value`), after which approve returns 200 and effect 204. And `cookieValue()` returns the **first** match, so a duplicate cookie shadows the genuine one (probe: `SHADOW` beat `GENUINE`).

"Limits carried into implementation" names only `Secure`, `HttpOnly`, host-only scope and `SameSite` — never the `__Host-` prefix, cookie tossing, or subdomain multisite. Meanwhile `includes/class-request-stash.php` already solved exactly this at `BINDING_COOKIE = '__Host-wp_sudo_stash_proof'`, whose docblock names cookie tossing and `COOKIE_DOMAIN` on subdomain multisite **by name**. Phase 27 is the handoff spec for Phase 29 and should not carry forward a weaker contract than existing in-repo prior art, uncited. Add `__Host-` and "the server must not adopt a binding it did not issue" to the Limits.

### D2 — the file-write arm does not demonstrate content binding

`/candidate-effect` receives `{ intent, target, digest }` — **no file content is ever transmitted** — and the "digest" is the literal `sha256:server-held-proposed-bytes` the server itself pinned at preflight. The descriptor check is string equality against a server-chosen constant. Only `/candidate-upload-effect` genuinely recomputes over received bytes. But "Selected copied-cookie observations" says Browser A "preflights the exact action, target, and **proposed-content digest** before sending the write," and the contract says the server "recomputes any content digest." The mutation table is honest (the File-descriptor kill point is a changed **target**); the prose presents both arms as equivalent. Say plainly: **file-write proves target binding, upload proves byte binding.**

### D3 — the 409 immutability evidence contradicts the documented per-intent design

The fixture holds exactly **one** global `candidateIntent` and one `uploadIntent`. Browser B gets 409 because there is only one record to overwrite. But the contract states: *"A second browser may create its own intent if otherwise authorized and the design does not rely on intent-ID secrecy."* Under that design B's preflight returns 200 with its own ID and the 409 never occurs — so "Browser B cannot overwrite Browser A's immutable prepared intent" is evidence for behaviour the production design explicitly will not have.

Related: running the full suite under `PHASE27_DISABLE_CANDIDATE_PREFLIGHT_IMMUTABILITY` fails **two** tests (#7 and #8), so the rate-limit test's `[409, 409, 429]` expectation is entangled with the same singleton artifact while the table names only #8's kill point. The genuinely load-bearing guards are approval-binding and effect-binding — scope the claim accordingly.

> **All four reviewers independently identified this singleton-intent artifact.** The upload reviewer additionally probed that a second *legitimate* upload — fresh preflight and fresh password re-approval, both 200 — still returns 409, and that `used` is a permanent global flag rather than per-intent consume state; the effect handler never *looks up* an intent by ID. The suite demonstrates intent **equality**, never intent **selection**.

### D4 — framing collision with the shipping plugin's docs

`main` states the analogous claim **without** the qualifier in at least four places — each true for `wordpress_logged_in_*`-only theft, false for a jar clone that also carries `wp_sudo_token`:

- `docs/security-model.md`, "Session Binding" section: "A stolen session cookie on a different browser … will not have a valid sudo session."
- `docs/security-model.md`, first threat bullet: "a stolen session cookie cannot perform covered gated actions… The sudo session is cryptographically bound to the browser."
- `docs/FAQ.md`, session-binding answer: "A stolen session cookie on a different browser will not have a valid sudo session."
- `docs/FAQ.md`, grace-period answer: "A stolen cookie on a different browser does not gain grace-period access."

`docs/FAQ.md`'s opening also says Sudo "is strong against a remote stolen or **cloned** cookie" — "cloned" reads as precisely the case this PR excludes. Two further hazards: the salt-degradation bullet in `security-model.md` ("the victim's existing `wp_sudo_token` cookie is *not* the barrier; the login session is") is correctly scoped to a DB-write attacker holding the auth salt but inverts this thesis if lifted out; and **both** mechanisms are called "binding" (`__Host-wp_sudo_stash_proof` stash binding vs. the Phase 27 approval binding), which readers will conflate. *(Being handled on #476.)*

### D5 / D6 — small

- "a full cookie-state clone can **race** to consume an approval" — there is no race in test 9; Browser A never submits, the clone simply redeems. Use "consume".
- Trace A step 3 and PROOF-03 both turn on B being able to "mint valid WP nonces". The fixture models **no** nonce or CSRF token anywhere; every endpoint accepts a bare same-origin fetch. The nonce clause is vacuously satisfied, not tested. Say so or drop it.

---

## Claim discipline — clean on four independent checks

Every reviewer was asked specifically to hunt for XSS or device-binding overreach. **None found any.**

- The record prints the **rejected** statement verbatim ("Sudo closes XSS-to-RCE while attacker-controlled JavaScript remains active…") and states that "Active script in Browser A can exercise its ambient server-side approval without reading a bearer. That is why active same-origin XSS is an explicit non-goal rather than a caveat hidden beneath HttpOnly." `REQUIREMENTS.md` TRUST-04 is the enforcing checkbox.
- Device binding is explicitly disclaimed: "another possession secret, not a hardware- or device-bound credential." Test 9 is the evidence for the concession.
- Atomicity is already correctly bounded to Phase 29.
- The upload flow is candidly a **single document** — file input, password, approve and submit all same-origin — so same-origin script would own it entirely, and the doc says so.

The one claim that overreaches is the accepted statement itself: "independently reaching the tested effects" is falsified by the unthrottled reauthentication path, which was executed end to end. It holds only if a throttled reauth step is assumed, and that assumption appears in neither the contract nor the fixture.

---

## Later-phase concerns — recorded, not for Phase 27 scope

- **Binding fixation** is the exploit form of D1: an attacker who can plant a cookie (subdomain cookie tossing, HTTP response injection on a sibling host, a cookie-injection bug) fixes A's binding to a known value and redeems A's approval holding only the auth cookie. Outside trace A's stated capabilities, but the cheapest conversion of "auth-cookie-only" into "full clone". Phase 29 should require server-side attestation — an HMAC over the session token, or a server-held issued-binding record.
- **Guards run after the whole body is buffered.** `await readBodyBuffer()` precedes every check, so an unauthenticated caller with no valid binding gets 64 MiB fully buffered before receiving 403 (probe-confirmed, 133 ms). The WordPress analogue is that PHP has already written the temp file before any gate runs — relevant to **where Phase 28 places the early effect veto**.
- **No failed-redemption budget.** Five consecutive wrong-digest submissions each return 409 without burning the approval; correct bytes then succeed. Correct fail-open-to-retry, but unbounded.
- **Approval binds the browser binding only.** `candidateIntent` has no actor or login-session-verifier field, though the contract claims approval stores "actor, login-session verifier, browser binding, immutable descriptor, and expiry". Expiry is disclaimed in Limits; actor and login-session verifier are not.
- **Availability tradeoff absent from Limits.** An attacker holding the copied auth cookie can burn the victim's 3-per-60s preflight budget and deny the gated action; the legitimate user's own preflight consumes budget.
- **`preflightAttempts` entries are overwritten but never deleted.** Bounded here (≤2 entries); under production session-token keying it grows one entry per session forever. The preflight budget also doubles as the intent-store growth bound, so an insider who resets it by re-login has unbounded intent-store growth.
- **`SameSite=Strict`** means a cross-site-initiated navigation into wp-admin carries no binding, so a fresh one is minted and any pending approval becomes unredeemable. Fails closed, but a UX cliff for Phase 30.
- **Binding issued to fully anonymous callers** on `GET /copied-cookie-upload-candidate` (probe-confirmed 200 + `Set-Cookie`). Harmless here since the binding alone authorizes nothing; production issuance policy is a Phase 28/29 decision.
- **Informed confirmation, for Phase 30.** The record openly concedes a server cannot derive a trusted package label before the bytes arrive and that client hashing is untrustworthy under active XSS. What a user actually approves is a generic server label plus a 64-hex digest they cannot verify against the zip they picked. Not overstated — but thin ground for "informed".
- **Test 9 survives all 16 mutations** (confirmed under three). It is a pure positive control carrying no information alone; its entire meaning comes from being test 8's differential partner. Fine as-is — but if anyone edits test 8, test 9 silently stops proving anything.

## Process

**The research suite runs in no CI workflow and has no npm script.** Neither `package.json` nor any file in `.github/workflows/` references `research.playwright.config.ts` or `tests/e2e/research/`; the E2E workflows enumerate specs under `tests/e2e/specs/` explicitly. The baseline is reproducible only by the hand-typed command in the record, and the 18 `PHASE27_DISABLE_*` toggles have no harness iterating them. Mutation runs being manual is normal; the **baseline** silently rotting is not. This matches the repo's known CI-integrity trap — a new spec not wired into the matrix never runs.

*One reviewer detected a sibling agent writing to a shared scratchpad path mid-run, discarded those results and re-ran everything in a uniquely-named private directory. The figures above are from the clean re-run.*

---

## Suggested disposition

**Before merge —** the U-1 assertion (a same-binding post-approval re-preflight returning 409), the D-1/D-2 mutation-table label corrections, the singleton-intent and transport-shape bullets in Limits, the `__Host-`/no-adopted-binding additions to the cookie contract, the file-vs-upload binding distinction, and the small `race`→`consume` and nonce-clause corrections.

**Carry into Phase 28/29 —** U-3, U-4, U-5, binding attestation, the probe-budget keying question with its lockout tension named as an open question rather than a settled property, and CI wiring for the baseline.

dknauss:
## Addendum — reproduction detail

The consolidated report above preserved guards, kill points, observed results and reviewer scope, but compressed the reproduction commands. Recording them here so re-review against a future SHA can re-run rather than re-derive. All figures in the parent comment were produced at `d3693b7`.

### Environment

Each reviewer independently confirmed `node_modules/` and `vendor/` in `/Users/dan-knauss/Code/Sudo-phase-27` are **real directories, not symlinks** (checked with `readlink`), so no cross-checkout `__DIR__` resolution. `@playwright/test` 1.62.0 resolves locally.

### Baseline

```
npx playwright test --config tests/e2e/research.playwright.config.ts --project=chromium
```

→ **10 passed**, 4.9–5.5 s across four independent runs. This is the record's "focused ordinary command" and it is accurate.

### Mutants

```
PHASE27_DISABLE_<NAME>=1 npx playwright test --config tests/e2e/research.playwright.config.ts --project=chromium --grep '<scope from the record>'
```

All **18** documented mutations were executed this way — 16 table rows plus the 2 negative-control mutations described in prose — and **18/18 were killed at the assertion the record names**.

A `diff` of the set of `PHASE27_*` names appearing in `tests/e2e/fixtures/phase27-research-server.ts` against the set appearing in the decision record comes back **identical**: no confabulated switches, no undocumented ones.

**Kill-point lines in `tests/e2e/research/phase27-handoff.spec.ts`**, all 16 distinct: 511, 524, 537, 550, 573, 587, 606, 793, 806, 819, 832, 851, 861, 884, 384, 426.

Upload mutants map to kill points as: `IMMUTABILITY`→793, `PASSWORD_CHECK`→806, `APPROVAL_BINDING`→819, `APPROVAL_INTENT`→832, `BINDING`→851, `DIGEST_CHECK`→861, `ONE_USE`→884.

**Individually observed mutant outputs:**

- `PHASE27_DISABLE_PREFLIGHT_CAPABILITY=1` → the anonymous caller receives `200` with the full body `{"digest":…,"id":"candidate-file-write-1","label":"Write sample-plugin/sample.php","target":…}` instead of `403 Forbidden`.
- `PHASE27_DISABLE_PREFLIGHT_RATE_LIMIT=1` → the fourth probe returns `409`, not `429`.
- `PHASE27_DISABLE_CANDIDATE_PREFLIGHT_IMMUTABILITY=1` → fails **two** tests (#7 and #8), not one. The record's table names only #8's kill point, which is how the singleton-intent entanglement with the rate-limit test's `[409, 409, 429]` expectation stays invisible.

### Fixture line references for the un-killed guards

In `tests/e2e/fixtures/phase27-research-server.ts`:

| Guard | Line(s) |
|---|---|
| `! authenticated` on `/candidate-upload-approve` / `/candidate-upload-effect` | 920, 956 |
| `binding === ''` on approve / effect | 921, 957 |
| `preflight.kind !== 'plugin'` | 870 |
| digest-shape regex `^[a-f0-9]{64}$` | 871 |
| `await readBodyBuffer()` (precedes all guards) | 948 |
| `uploadIntent.used` check | 971 |
| `uploadIntent.used = true` write | 998 |
| effect-stage intent+digest bundled condition | 985–989 |

The one-use section is atomic because there is no `await` between lines 971 and 998 and the hash is synchronous — Node's run-to-completion. That reason does not transfer to php-fpm.

`authorizePreflight()` builds its budget key as `const rateKey = scope + ':' + auth;` where `auth` is `cookieValue( request, 'wp_auth' )`.

### Probe methodology

Findings that required behaviour the spec does not exercise were produced with throwaway harnesses **outside the repo**, never by editing tracked files:

- Raw `fetch` drivers (the adversary is a script holding a cookie, not a browser) for the unthrottled-approval and rate-key-rotation results.
- The fixture imported by absolute path under Node 24 type-stripping — it is pure `node:http` plus inline HTML with no path-relative dependencies, so this exercises the actual shipped code.
- Un-killed-guard experiments ran against **copies** of fixture + spec + config in a private scratchpad directory, verified to reproduce 10/10 baseline before any clause was deleted.

`git status` was empty and HEAD unchanged before and after every run; Playwright artifacts are gitignored.

**One contamination note, recorded because it affected results:** one reviewer detected a sibling agent writing to the shared scratchpad path mid-run, discarded those results, and re-ran everything in a uniquely-named private directory. All figures reported are from the clean re-run.

### Measurements quoted in the parent comment

- Timing-oracle probe, 2 MB body, medians: unauthorized-403 **7.1 ms**, authorized-unknown-target-403 **8.2 ms**, authorized-known-target-200 **7.8 ms**.
- Rate-key rotation: fixed session token, 6 probes → `[200,200,200,429,429,429]`; rotating token per request → `[429,200,200,200,200,200]`.
- Unthrottled approval: **500** sequential wrong-password POSTs to `/candidate-approve`, all `403`, no delay; then correct password `200`; then `/candidate-effect` `204`.
- 50 binding-less requests → all `403`, zero budget consumed (next 4 bound probes still `[200,200,200,429]`).
- U-1 post-approval byte swap: approved digest `17f75876c4a7e94e98d23d54290a11c43d386ea46977a387b6aa249a2e930b01`, executed digest `407efd44…`, status `204`, suite still 10/10 green.
- Upload byte fidelity: SHA-256 of `50 4b 03 04 00 ff 80 41` = `17f75876…` (matches spec:779); a UTF-8 round-trip would give 12 bytes `504b030400efbfbdefbfbd41` and digest `3a21911c…`.
- Concurrency: 2, 8 and 32 simultaneous redemptions → exactly one `204` and one observation every time.
- Body buffering: unauthenticated caller with no valid binding, 64 MiB fully buffered before `403` (133 ms).

### Severity note carried forward

The unthrottled-approval finding was classified **reachable bypass** by the preflight reviewer (executed end to end) and **later-phase concern** by the cookie-state reviewer (reasoning it is reachable by a full-cookie-state clone rather than within trace A's stated capabilities). Both endpoints — `/candidate-approve` and `/candidate-upload-approve` — lack an attempt budget. The split is a genuine scope disagreement and is left unresolved rather than averaged.

dknauss:
## Current-main restack and architecture checkpoint

Rebased the full Phase 27 evidence tranche onto `origin/main` at `9d29946` and regenerated metrics rather than resolving the generated file by hand. New head: `250ee7a`.

Fresh local evidence on the rebased tree:

- Chromium baseline: 15/15 passed.
- Node mutation manifest: 15/15 guards killed first at their named assertions.
- Real WordPress 7.0.2 / PHP 8.2 baseline: 6/6 passed.
- WordPress mutation manifest: 6/6 guards killed first at their named assertions, including atomic consumption, effect input, lifecycle revocation, and fail-closed storage.
- PHPUnit: 1,308 tests / 3,907 assertions.
- PHPCS, PHPStan level 6, Psalm, i18n, metrics, and 69-source registry verification passed. Existing source drift warnings reproduce independently; no source check failed.

The architecture claim remains intentionally narrow. This PR supplies evidence for an action-bound candidate against copied authentication-cookie use; it does not establish a production/core implementation, a real wp-admin preflight adapter, multisite behavior, active same-origin XSS resistance, or protection against cloning complete browser cookie state.

That distinction now blocks treating #537 disabled controls as the security architecture. #537 may be a compatibility UX adapter under the existing reusable sudo window, or a later real adapter for this action-bound candidate, but those are different products and must be chosen explicitly.

dknauss:
Closing as superseded by the final research finding in PR #545. The action-bound approval candidate and its accepted boundary informed the constructive Core direction, but it is not a production successor architecture. This PR remains part of the research history.

Issue #456, opened 2026-07-27
Auto-merge fires on thread resolution, so PRs can merge with review fixes unapplied


dknauss:
Hit **twice on 2026-07-27**, once with real consequences. Filing so it survives the sessions that found it.

## The trap

With `--auto` armed and all checks green, **resolving the last review thread is the merge trigger.** Resolution is a click. It verifies nothing about whether the fix the thread asked for exists.

So the natural order — reply to threads, resolve them, then push the fixes — merges the PR at step 2.

### Occurrence 1: #444 merged with two P1 fixes unapplied

Five threads were dispositioned and resolved; auto-merge fired; `3e1a35b` landed **without** the changes two of those threads required. `docs/developer-reference.md` shipped a sentence stating the opposite of runtime behaviour (`new_admin_email` described as not a recognised target, after the same PR added it to `TARGET_PARAMS`), and `docs/release-status.md` — canonical for release state — had no record of `options.critical` becoming non-replayable.

Caught only because the merged tree was checked directly rather than the resolutions trusted. Repaired in #451.

### Occurrence 2: the orphan branch

The author pushed immediately after resolving. By then the merge had completed and **GitHub had deleted the branch**, so the push *recreated* it — a branch attached to no PR, containing the only copy of the fixes. The sole signal was `* [new branch]` in the push output.

A session that resolved threads and ended its turn would have left its work stranded on a branch pointing at nothing, with every surface reporting the PR merged successfully.

## Rule

**Push first. Resolve second.** Never resolve a thread whose fix is not already on the remote.

If a thread is being resolved as *deferred* rather than *fixed*, that is fine — but say so on the thread, because the next reader cannot distinguish "resolved because fixed" from "resolved because deferred" from the resolution alone.

## Suggested guard

Reviewer/merge instructions should require verifying the **merged tree**, not the thread state: after a PR merges, `git show origin/main:<path>` for each file a review finding named. That is what caught occurrence 1.

## Class

Same shape as the other findings this release, and the reason it is worth an issue rather than a footnote: **a signal that reports done without checking.** Cf. #434, #442, #447 (registry exemption keyed on identity), and the pre-commit hook that was never installed (#427) — mechanisms whose green means something narrower than it appears to.

dknauss:
Closing as not planned because WP Sudo has concluded as a research prototype and will be archived read-only. This issue remains available as historical context, but there is no continuing product backlog.

PR #476, opened 2026-07-28
docs: qualify the session-binding claim — auth cookie alone, not complete cookie state

dknauss:
Seven user-facing claims said a stolen cookie on a different browser cannot use a sudo session, without saying **which** cookie. The mechanism does not support the unqualified version.

## The mechanism

A gated request is admitted when two things line up: the proof entry keyed by `hash( 'sha256', wp_get_session_token() )` — derived from the auth cookie — and the `wp_sudo_token` cookie whose hash that entry stores.

- **Auth cookie alone:** attacker has the first, not the second. Refused. This is what the binding is for.
- **Complete cookie state:** attacker has **both**. Same verifier, matching token hash, valid HMAC. Nothing in the design distinguishes that browser from the original.

## Why this needed fixing

The internal implementation spec already conceded it — "does not stop a session-riding XSS or a **full-cookie-jar theft** during the window". `docs/security-model.md`, `docs/FAQ.md` and `readme.txt` did not, and `grep -rn "cookie jar"` across the docs returned nothing.

So the honest documents were the internal ones. For a project whose public claims are the ones under scrutiny — and which was just reclassified as a research prototype precisely to stop overclaiming — that is the wrong way round.

## What changed

Stated **once** in `security-model.md` as a named boundary; the other call sites qualify and link to it rather than restating, per the prose rule against a seventh copy of a claim.

Two framings worth flagging for review:

- **It avoids implying that cookieness is the flaw.** Both values are bearer secrets with essentially the same browser exportability, so the binding adds a *second instance of the same class*, not a different class. What it narrows is the attack **path** — a partial exfiltration reaching one value and not the other — not the attacker's capability. Any route that exports complete cookie state exports both.
- **`HttpOnly` is doing less work than it appears.** It stops script *reading* the cookie; it does not stop script *causing the browser to send* it. That is why an XSS active during the sudo window is listed separately as undefended: the proof is used, not stolen.

The defended case is therefore partial exfiltration — a value from a log, a proxy, a referrer leak. The undefended case is anything yielding complete cookie state: malware, a copied browser profile, physical access, a debugging bridge. Those imply a compromise broader than this plugin's boundary, but the claim should not read as covering them.

## Provenance

Prompted by the Phase 27 research branch adding a cloned-cookie-state negative control, so the research evidence and the public documentation now describe the same boundary rather than diverging.

## Validation

Docs-only (3 text files; the hook took the docs-only path). `composer verify:sources` 57 citations · `composer verify:metrics` in sync · `git diff --check` clean · the new `security-model.md` anchor resolves from the FAQ link.


dknauss:
## Independent review of the staged changes — verdict: REJECT (with a fast path to green)

A Claude subagent reviewed the four **staged** files for this PR, reading each via `git show :<path>` rather than the worktree. Tree left exactly as found; no `reviewer-approved` flag written.

The central security argument is **correct and independently confirmed**. Three in-scope violations of explicit project MUSTs block the commit, and they are small.

### Environment integrity

`vendor/` in the worktree is a real directory, not a symlink, and the classmap resolves `WP_Sudo\Gate` to this worktree's `includes/class-gate.php`.

### Mechanical checks — all green

| Check | Result |
|---|---|
| `composer test:unit` | OK — 1276 tests, 3827 assertions |
| `composer lint` | 24/24, no violations |
| `composer analyse:phpstan` | `[OK] No errors` |
| `composer verify:sources` | 57 citations checked, all snippets present |
| `composer verify:metrics` | in sync |
| `bash bin/make-pot.sh` → `git diff --stat languages/wp-sudo.pot` | **empty** — genuinely regenerated, not hand-patched |

### Verified against live source

Fetched `wordpress-develop` trunk directly — not the registry, not model memory:

- **`wp_set_auth_cookie()`** (`src/wp-includes/pluggable.php`): `$token = $manager->create( $expiration );` once, then `wp_generate_auth_cookie( …, $scheme, $token )` and `wp_generate_auth_cookie( …, 'logged_in', $token )`. **One token, both cookies — claim correct.** Both `setcookie()` calls pass `true` for httponly, so the HttpOnly premise in the summary blockquote is correct too.
- **`wp_parse_auth_cookie()`**: `explode( '|', … )`, `count !== 4` reject, `list(...)`, `compact(...)`. **No HMAC check, no expiry check — claim correct.**
- **`wp_get_session_token()`** (`src/wp-includes/user.php`): `wp_parse_auth_cookie( '', 'logged_in' )`. **Verifier comes from `LOGGED_IN_COOKIE` — claim correct.**

**The item-3 retraction is sound.** An attacker holding only the admin auth cookie does hold the token and can attach a four-field junk `LOGGED_IN_COOKIE` to key the lookup — `wp_validate_auth_cookie` at `determine_current_user` priority 10 authenticates off the auth cookie in wp-admin and short-circuits `wp_validate_logged_in_cookie`, so the fabricated cookie does not break authentication. The earlier "cannot even key the lookup" draft was wrong; the retraction corrects it.

Also confirmed: the retracted phrase **"the binding is itself a cookie" survives nowhere** in the repo (swept `.md`, `.txt`, `.php`, `.js`, `.pot`, `.json`, excluding `.git`/`vendor`); `readme.txt:72` already carries corrected framing. Both registry IDs exist. The help-tab string is a single `__()` call with the apostrophe correctly escaped for the single-quoted PHP string, and the `.pot` line ref matches the staged file. No sentence claims defense against XSS generally or device binding.

---

## Blocking findings

### A — MAJOR: a false claim about this repo's own code, built on a scope word

`docs/security-model.md`, "HTTPS and reauthentication replay (#322)" bullet:

> `Challenge::build_replay_response_data()` assigns `replay_disabled` unconditionally

The assignment is immediately followed by a conditional overwrite:

```php
$refusal_reason = 'replay_disabled';

if ( ! $credential_verified ) {
    $refusal_reason = 'no_credential_this_request';
}
```

`replay_disabled` is the outcome **only** when a credential was verified on this request. The `wp_sudo_replay_refused` docblock and `docs/developer-reference.md:675` document both values, and the doc's own parenthetical also runs from `complete_active_session_request()` / `render_resume_page()`, where the reason is always the other one. This is AGENTS.md's named pattern verbatim: a true narrow observation inflated by a scope word, borrowing credibility from the correct clause beside it. `CHANGELOG.md:88` already has accurate phrasing — *"has no branch that can return a replay instruction, which is the whole guarantee and is verifiable by reading one method."*

*(Note: this claim originated in a session handoff and propagated into the review brief as well as the doc. The underlying security guarantee holds; only the supporting detail was wrong.)*

### B — MAJOR: unregistered third-party claim (AGENTS.md MUST)

The retraction's load-bearing premise is a **new** claim about WordPress core with no registry ID:

> `wp_set_auth_cookie()` mints a single session token and puts it in **both** cookies

`GB-SESSION-TOKEN-SRC` covers `wp_get_session_token()`; `GB-AUTH-COOKIE-PARSE` covers `wp_parse_auth_cookie()`; `GB-LOGGEDIN-COOKIE-PATH` cites `wp_set_auth_cookie()` but for cookie *paths*. The claim is **true** — but nothing mechanical will catch it drifting, which is what the registry exists for. Needs a row anchored on `$token = $manager->create( $expiration );` or one of the two `wp_generate_auth_cookie(...)` lines.

### C — MAJOR: the two staged files contradict each other on replay

- **`docs/security-model.md` (staged):** "Since #322 nothing is resumed at all … no branch that resumes an intercepted action. The cloned-jar boundary below therefore concerns the **sudo session**, not replay."
- **`docs/FAQ.md` (staged), line 9 — the very sentence this PR edits:** "a stashed action is never resumed automatically **unless** the browser presenting the reauthentication is the one that created it, and the challenge now names the concrete target".

The FAQ describes the superseded #322-**v2** origin-bound-replay design as live behaviour. It is contradicted by the code, by `CHANGELOG.md:88`, by `docs/release-status.md:23` ("Automatic replay is **removed rather than conditioned**"), and by `docs/developer-reference.md:569`. `docs/FAQ.md:106` elaborates the same dead design at length, and `:108` self-declares as *"the canonical description of when a request is and is not resumed"* — with `readme.md`, `readme.txt` and `docs/two-factor-integration.md` pointed at it.

The staleness predates this PR. What makes it blocking here is that the commit **adds the assertion that refutes it, in a sibling staged file, to defend the very scope split this PR turns on** — and leaves it standing in the one sentence the author had open. A reader of FAQ line 9 applies the new cloned-jar caveat to a browser-bound replay that does not exist and reaches the opposite of what `security-model.md` now says.

### D — MINOR: ambiguous antecedent

"…nor against a copy of the browser's complete cookie state, **which carries the sudo binding too**." There are two bindings in this model — the sudo-session `wp_sudo_token` cookie and the now-inert `__Host-` stash binding — and the clause immediately preceding it concerns the *replay* binding. Given this PR's `security-model.md` edit exists specifically to say the boundary concerns the sudo session and **not** replay, the FAQ should name it: "the sudo session's binding cookie". On its strict reading the sentence is accurate — it is the ambiguity, not the claim, that is the defect.

### E — MINOR: redundant leftover

"What stops them is the second conjunct — they do not have `wp_sudo_token`. An attacker holding only the auth cookie has the first and not the second, which is exactly what the binding is for." The second sentence restates the first, and the paragraph is unreflowed mid-edit.

### F — INFO: pre-existing stale code comments

`Challenge::replay_stash()`'s docblock still describes browser JS that "Redirects for GET requests / Builds and submits a hidden form for POST requests", and the closing comment block in `build_replay_response_data()` says "#322 v2 (origin-bound replay) restores seamless auto-replay for the same-browser case". Neither is true of shipped behaviour.

---

## Cross-referenced from #470

The cookie-state reviewer on #470 independently found that `main` states the analogous session-binding claim **without** the qualifier in four public places — each true for logged-in-cookie-only theft, false for a jar clone that also carries `wp_sudo_token`:

- `docs/security-model.md`, "Session Binding" section
- `docs/security-model.md`, first threat bullet
- `docs/FAQ.md`, session-binding answer
- `docs/FAQ.md`, grace-period answer

Plus `docs/FAQ.md`'s opening describing Sudo as strong against a "cloned" cookie — precisely the excluded case. Two further hazards: the salt-degradation bullet in `security-model.md` inverts the thesis if lifted out of its DB-write-attacker context, and **both** mechanisms are called "binding", which readers will conflate.

---

## Agreed resolution

Taking the broader coherent fix rather than the minimal one — leaving the unqualified claims in place would make this PR accurate at its edited lines while the repository's main security model and FAQ continued to overstate the same protection. This is one logical documentation change: accurately describe current replay behaviour, and the boundary between auth-cookie-only theft and full-cookie-jar cloning.

1. Replace "assigns `replay_disabled` unconditionally" with the narrower verified fact — no branch of `Challenge::build_replay_response_data()` can return a replay instruction.
2. Add a verified upstream-source registry entry for `wp_set_auth_cookie()` minting one session token used for both cookies.
3. Rewrite the FAQ opening clause: current Sudo does not automatically resume retained requests at all; the origin-bound auto-replay design was **removed, not made conditional**.
4. Correct the canonical replay explanation at FAQ 106–108 **in this PR** — fixing only the opening sentence while leaving the linked canonical answer false would not resolve the contradiction.
5. Qualify all four public session-binding statements: protection applies when the WordPress authentication cookie is stolen alone; it does not distinguish an attacker who also clones the Sudo binding cookie; and name that cookie explicitly rather than saying "the binding".
6. Sweep the repository for `stolen cookie`, `cloned cookie`, `session binding`, `resume`, `replay` and similar wording so no other public duplicate is missed.
7. File the stale PHP docblocks (F) separately unless the corrected prose requires touching their code context.

**#470's proposed protocol is deliberately not imported here.** This PR describes the released plugin exactly as its current code behaves; #470 is reopened research and is not authoritative evidence for a future mechanism.

A fresh review will then be requested against the final SHA with three explicit questions: does any public document still imply automatic replay exists; does any statement still imply full-cookie-jar cloning is stopped; and is every load-bearing WordPress-core claim registered and mechanically verified.

dknauss:
## Correction to the resolution plan above — three of the four claims were already fixed on this branch

Step 5 of the agreed resolution says to "qualify all four public session-binding statements". **That premise is stale, and I should correct it before anyone acts on it.**

The four unqualified statements were found by the cookie-state reviewer on #470, which read **`origin/main`** — not this branch. Commit `38c8afa` on `docs/cookie-jar-boundary` had already qualified them. Verified with `git diff origin/main...HEAD`:

| Location | State on `main` | State on this branch |
|---|---|---|
| `docs/FAQ.md`, session-binding answer | unqualified | **already qualified** — "the attacker lacks the second cookie. A copy of the whole cookie jar carries both and is **not** covered" |
| `docs/FAQ.md`, grace-period answer | unqualified | **already qualified** — "(a cloned cookie jar would — same boundary as the sudo window itself)" |
| `docs/security-model.md`, first threat bullet | unqualified | **already qualified** — "so this holds against theft of the auth cookie alone and not against a cloned cookie jar" |
| `docs/security-model.md`, "Session Binding" section | unqualified | **already qualified** — "A stolen **authentication** cookie on a different browser…" |

`readme.txt` was also already updated by the same commit, in two places.

So they are unqualified on `main` precisely because **this PR is what fixes them**. The reviewer's finding was correct about `main` and would have been correct about the repository as a whole had this PR not existed; it should not be read as four outstanding edits.

### What genuinely remained, and is now done

1. **Naming the cookie explicitly** — Codex's third sub-point was the part not yet satisfied. The branch said "the second cookie", "a cloned cookie jar", "by a second cookie" without ever naming `wp_sudo_token`. Now named at each site.
2. **A fifth location nobody had flagged** — `docs/developer-reference.md`, in the `is_within_grace()` reference: *"A stolen cookie on a different browser, or a forged proof, does not gain grace access."* Unqualified, not touched by commit `38c8afa`, and outside the four the #470 reviewer named. Found by the repo-wide sweep in step 6.

### Sweep scope

Swept `stolen cookie`, `cloned cookie`, `cookie jar`, `session binding`, `resume`, `replay`, `auto-replay` and `resumed automatically` across `.md`, `.txt`, `.php` and `.js`, excluding `vendor/`, `node_modules/` and `.git/`. Matches in `.planning/`, `tests/`, `poc/` and `docs/archive/` were reviewed and deliberately left: they are internal working notes and historical records, not public documentation, and several are accurate statements *about the superseded design as history*. `CHANGELOG.md` likewise describes the 4.9.0 change correctly and needs no edit.

The replay corrections (steps 1, 3 and 4) stand as agreed and are unaffected by this correction.

dknauss:
## Resolution — final SHA `03fd142`

**No inline thread is resolved by this comment.** It records which findings the pushed commit addresses, so the threads can be closed on their merits by a human rather than to unblock a merge (#456's failure mode).

Approved staged tree `aeaede379246f51a7d7c3959e81712eae140313a`; committed tree verified identical; local HEAD, the remote ref and this PR's `headRefOid` all agree on `03fd14295f0843db6e7016640008e026a3d2b658`.

### The four Codex threads

| Thread | Addressed at `03fd142` |
|---|---|
| **Remove XSS from the cookie-exfiltration examples** | Done, then **narrowed further**. The blockquote in `docs/security-model.md` → *What It Protects Against* no longer says script cannot read the cookie. It states core sets auth cookies `HttpOnly` **by default**, then names three limits: `wp_set_auth_cookie()` is pluggable, the `send_auth_cookies` filter can suppress the path, and script that cannot *read* the cookie can still **ride** the session — undefended. |
| **Propagate the boundary to the live help tab** | Done. `Admin::add_help_tabs()`'s "What Sudo is strong at" entry now distinguishes auth-cookie-only theft from a complete-cookie-state copy. Single `__()`, apostrophe escaped, `.pot` regenerated and matching. |
| **Name the logged-in cookie that supplies the verifier** | Done, and the reasoning was **reversed** during review. An earlier draft claimed an attacker "cannot even key the lookup"; that was wrong. `wp_set_auth_cookie()` mints one token and puts it in both cookies, and `wp_parse_auth_cookie()` splits without verifying an HMAC or expiry — so an auth-cookie holder can fabricate a `LOGGED_IN_COOKIE` that keys the lookup exactly. What stops them is not holding `wp_sudo_token`. Registered as `GB-AUTH-COOKIE-ONE-TOKEN`. |
| **Do not extend the cloned-jar caveat to disabled replay** | Done. The supporting detail was also **corrected**: the claim that `build_replay_response_data()` assigns `replay_disabled` *unconditionally* is false — a conditional overwrite to `no_credential_this_request` follows it. Replaced with the narrower verified fact that no branch of the method can return a replay instruction. |

### Findings from the review comments above

Linking back: [four-agent review of #470](https://github.com/dknauss/Sudo/pull/470#issuecomment-5100156995), [reproduction addendum](https://github.com/dknauss/Sudo/pull/470#issuecomment-5100187784), [this PR's first review](https://github.com/dknauss/Sudo/pull/476#issuecomment-5100161679), [correction to its resolution plan](https://github.com/dknauss/Sudo/pull/476#issuecomment-5100190757). Those four are preserved unedited as the audit record.

- **Four unqualified session-binding claims** reported by #470's cookie-state reviewer — already qualified by `38c8afa` on this branch before that review ran; it read `origin/main`. `03fd142` adds what was genuinely missing: `wp_sudo_token` named explicitly at each site, plus a fifth location nobody had flagged, `docs/developer-reference.md`'s `is_within_grace()` reference.
- **Public replay staleness** — corrected across `docs/FAQ.md`, `docs/security-model.md`, `docs/developer-reference.md`, `readme.txt` (including the Upgrade Notice WordPress shows in the update UI), `docs/two-factor-integration.md`, `docs/two-factor-ecosystem.md`, `docs/two-factor-authentication-flow.md`.
- **A fabricated Site Health test name** ("Sudo Reauthentication Replay") removed; `Site_Health` registers no such test.
- **New core claims registered**: `GB-AUTH-COOKIE-ONE-TOKEN` (pluggable.php:1137) and `GB-AUTH-COOKIE-HTTPONLY` (:1193), both verified against `wordpress-develop` trunk. The first deliberately pins the *sharing* of `$token` rather than the mint, because the mint line is not unique in that file.

### Deliberately not in this PR

The in-code comment debt — **255 matches, ~78 stale claims** — is inventoried in **#478** and **#479**. Five review rounds here each found the same defect one layer further out, which is what made this PR unbounded. Highest-severity item carried over: `Gate::evaluate_diagnostic_request()` states *"Since #322 replay **is** authorised by a conjunction this method cannot evaluate"*, which contradicts the `docs/FAQ.md` paragraph corrected in this very commit.

Also recorded there, and newly found by the final review: **`docs/release-status.md` L100** describes the retired eligibility architecture as current ("`options.critical` is **now** non-replayable", "was replay-eligible") while L23 and L166 of the same file are correct — an internal inconsistency in the repo's canonical release-state source.

dknauss:
## Resolution of all review threads — final head `f22dd0d`

Recording what addressed each thread before resolving them, so they close on their merits rather than to unblock the merge (#456's failure mode).

### The four original threads

| Thread | Addressed |
|---|---|
| **Remove XSS from the cookie-exfiltration examples** | Done, then narrowed twice more. The blockquote now states core sets auth cookies `HttpOnly` **by default**, and names three limits: `wp_set_auth_cookie()` is pluggable, the `send_auth_cookies` filter can suppress it, and script that cannot *read* the cookie can still **ride** the session. A later pass caught that the replacement example — "a referrer leak" — was itself wrong: a `Referer` carries a URL, not a `Cookie` header. |
| **Propagate the boundary to the live help tab** | Done. `Admin::add_help_tabs()` now distinguishes auth-cookie-only theft from a complete-cookie-state copy. |
| **Name the logged-in cookie that supplies the verifier** | Done, and the reasoning was **reversed** mid-review. An earlier draft claimed an attacker "cannot even key the lookup". False: `wp_set_auth_cookie()` puts one token in both cookies and `wp_parse_auth_cookie()` splits without verifying HMAC or expiry, so the lookup *can* be keyed. What stops them is not holding `wp_sudo_token`. Registered as `GB-AUTH-COOKIE-ONE-TOKEN`. |
| **Do not extend the cloned-jar caveat to disabled replay** | Done. The supporting detail was also corrected: the claim that `build_replay_response_data()` assigns `replay_disabled` *unconditionally* was false — a conditional overwrite to `no_credential_this_request` follows it. Replaced with the verifiable narrower fact that no branch can return a replay instruction. |

### Later threads

- **`post_mode: none` eliminates retention** — false. `capture_target()` runs independently of the body policy and reads `TARGET_PARAMS` straight from the superglobals, so a rule carrying `plugin`/`user_id`/`file` still stores that value, by design.
- **Limit the cloned-cookie caveat to a live sudo window** — added, then corrected again: a first attempt claimed a jar copied "before the victim ever ran a challenge" carries no proof. False, and understating risk — `wp_login` auto-grants a session by default, so a jar copied moments after an ordinary **password-only** login is the *most* dangerous copy.
- **Preserve the handler-endpoint landing caveat** — added, then swept to **seven** sites after the first attempt fixed one and left six siblings.
- **Remove referrers from the cookie-exfiltration examples** — done, see above.
- **Stop claiming POST allowlists feed the challenge** — done. `describe_stash_target()` never reads the stored body; with replay gone an allowlist retains data for no benefit, so `post_mode => 'none'` is now recommended absent a demonstrated consumer.
- **Qualify the remaining grace-window binding claims** — done at all three sites (`GRACE_SECONDS` docblock, `is_within_grace()` docblock, `CHANGELOG.md:593`).
- **Update runnable checks for removed replay** — done for `tests/MANUAL-TESTING.md`, which **no PR had touched**: eleven instructions expected a replayed action, a `replayed` record a dormant hook cannot emit, or a Cancel returning to the originating page when `cancelUrl` is forced server-side to the dashboard. The other two files named are corrected in #484.
- **Qualify secret redaction as POST-only** — done, and this was a real security misstatement: `build_original_url()` stores the full `REQUEST_URI`, so a GET-gated rule keeps a URL secret in the transient.
- **Do not promise atomic one-shot hook delivery** — done. `get()`/`delete()` are unlocked transient wrappers, so concurrent completions can both fire `wp_sudo_replay_refused`.
- **Limit the post-login clone example to persistent sessions** — done, qualified to password-only logins.
- **Remove the false referer dependency** — done. The comment claimed `_wp_http_referer` is kept in the stash allowlist "so the challenge can land the user back on the right tab". Nothing reads a stashed `_wp_http_referer`; both `options.wp_sudo` topologies are diverted to the neutral page anyway; the tab is preserved by the user's own live POST. The allowlist entry is vestigial.

### Filed rather than fixed

- **Record the new upstream verification in the commit message.** Both new core claims *are* registered and mechanically verified — `GB-AUTH-COOKIE-ONE-TOKEN` (pluggable.php:1137) and `GB-AUTH-COOKIE-HTTPONLY` (:1193) — and were re-checked against live `wordpress-develop` trunk by four separate passes. The request is for provenance in a **pushed commit message**, which cannot be corrected without rewriting history on a branch under review.
- **`CHANGELOG.md` lines 88 / 112 disagree about replay conditionality.** Pre-existing on `main`, untouched by this branch.

### Verification at `f22dd0d`

`composer test:unit` 1276 tests / 3827 assertions · lint 24/24 · PHPStan clean · `verify:sources` 59 citations · `verify:metrics` in sync · `.pot` regenerates byte-identical from a `git archive` extract of the staged tree. Every commit's tree was verified against its approval before pushing.

Issue #460, opened 2026-07-27
Unit suite is intermittently non-deterministic — varying assertion counts on identical input


dknauss:
Found by an independent reviewer during the #459 audit, on a **verified-clean tree** at `4b4f268`. Reported here because an intermittently red suite is corrosive in a specific way: it trains people to re-run rather than investigate, which is how a real regression gets waved through.

## Symptom

Three consecutive runs on identical input produced three *different* reds, then green:

```
Tests: 1273, Assertions: 3802, Errors: 1, Failures: 3.
Tests: 1273, Assertions: 3804, Failures: 4.
Tests: 1273, Assertions: 3808, Failures: 4.
OK (1273 tests, 3812 assertions)   ← and 25+ green runs since
```

**The varying assertion count is the interesting part.** Test count is stable at 1273, so the same tests run — but a different number of assertions executes each time, which means tests are taking different paths run-to-run.

## What is known and not known

- Both failure clusters **correlated with concurrent git/composer activity** against the repo.
- The reviewer could **not** reproduce it deliberately: not under heavy I/O, not with two concurrent `phpunit` processes, not with a concurrent `git worktree add`.
- **The failing test names were not captured** before it stopped reproducing. That is the gap in this report, and it is stated rather than dressed up.

An earlier occurrence in the same review was attributed to a concurrent agent mutating a shared worktree. **That explanation does not cover this one** — the tree was clean and verified, and the agent had been stopped.

## Why it matters beyond the annoyance

A single green CI run is weaker evidence than it looks. During this audit the reviewer stopped trusting whole-suite reds and re-verified every mutant with `--filter` against the specific guard test, recording *which* test failed — because "the suite went red" and "the guard caught it" are different claims, and a flaky suite makes them indistinguishable.

That is a sound practice regardless, but here it was forced.

## Suggested first steps

1. Run the suite in a loop (50–100×) on an otherwise idle machine and capture failing test names plus `--order-by=defects` output.
2. Check for state leaking between tests via the static caches the suite resets — `Action_Registry::reset_cache()` and friends — and anything reading real time or the filesystem.
3. Consider `--order-by=random` with a recorded seed in CI, so ordering dependence surfaces deliberately rather than by luck.

Not release-blocking: it does not contradict the green CI on #459, and every guard that release depends on was verified individually rather than through a whole-suite result.

dknauss:
[...]

PR #477, opened 2026-07-28
chore(460): instrument suite determinism, and reproduce the reported signature

dknauss:
Answers #460 rather than closing it by re-running until green.

## The report's inference does not hold — and the numbers point the other way

#460 recorded three consecutive reds with three different assertion totals on a clean tree, read as evidence that identical tests took different paths.

The totals were **3802 / 3804 / 3808 against a green 3812 — all below it.** A failed assertion throws (`Assert::fail()` → `AssertionFailedError`), so the rest of that method never runs. Uniformly-below is exactly what failure-abort produces, so it is *expected under* failure-abort and is not evidence for divergent paths.

Stated carefully: this does not rule divergence out. A skipped branch also reads uniformly-below. It removes the totals as evidence *for* it.

## The signature reproduces on demand — through dependency contamination

| condition | result |
|---|---|
| **isolated** worktree, own `vendor/`, verified autoload root | **43 runs, zero failures**, identical totals, one root |
| `vendor/` symlinked to a second checkout, that checkout's `includes/` changing mid-run | **8 runs → 3 green, 5 red**, three different failure sets, two different assertion totals |

The second row is on a tree `git status` calls clean: stable test count, varying assertions, intermittent green, correlated with concurrent activity. That is #460's report.

The mechanism is PHP resolving symlinks in `__DIR__`, so a symlinked `vendor/` makes Composer compute `$baseDir` in the other checkout and load every class from there. The same trap caught two agents independently on 2026-07-27 — once while verifying a guard test, once while a reviewer reproduced that result.

**This is a sufficient mechanism, not proof of what happened at `4b4f268`** — a symlinked `vendor/` leaves no trace afterward. It is the only hypothesis found that produces varying failures on a clean tree, and it matches the reported correlation with concurrent git/composer activity.

## The harness had the defect it exists to prevent

Worth stating plainly, because review caught it and it is the whole point of the tool. An earlier cut printed `ok` twice and **"No failures in 2 runs"** while PHPUnit never executed — paths were repo-relative with no `cd`, and status came from whether a failure line appeared rather than from the exit code.

Both fixed. Status now also requires a parsed test count, because **PHPUnit 9.6 exits 0 on an empty suite** (`No tests executed!`, `failOnEmptyTestSuite` unset) — the exit code alone would have re-manufactured the silent green and falsified the header's own guarantee.

Review additionally caught: invalid JSON on exactly the runs that matter (every test name carries `WP_Sudo\Tests\Unit\` backslashes), a confounder field destroyed by stack traces when the confounder was present, and a header claiming two fields the script never recorded.

## Relationship to the merged gate

[#472](https://github.com/dknauss/Sudo/pull/472) landed the commit-time defence — the pre-commit hook refuses a `vendor/` that is not the worktree's own. This covers the ad-hoc runs that hook cannot see, and makes the autoload root part of every recorded result so the question is never open again.

Also adds `.tmp/` to the **tracked** `.gitignore`; it was ignored only via `.git/info/exclude`, which is machine-local, and committed tooling writes there.

## Validation

`composer test:unit` 1276 / 3827 · `lint` 24/24 · PHPStan clean · `verify:metrics` and `verify:sources` in sync · `bash -n` clean. Reviewer-approved against staged tree `1421e188`, verified by extracting the blob with `git show :<path>` rather than reading the worktree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,8 @@ test-results/
 
 # claude-code-reviewer approval flag
 reviewer-approved
+
+# Scratch output from bin/suite-determinism.sh and the WP test-suite installer.
+# Previously ignored only via .git/info/exclude, which is machine-local — committed
+# tooling writes here, so the rule belongs in the tracked file.
+.tmp/
--- a/bin/suite-determinism.sh
+++ b/bin/suite-determinism.sh
@@ -1,0 +1,199 @@
+#!/usr/bin/env bash
+#
+# Run the unit suite repeatedly and record what a bare pass/fail line does not.
+#
+# #460 reports three consecutive reds with three different assertion totals on a
+# clean tree, read as evidence that identical tests took different paths. That
+# inference does not hold on its own: a failed assertion throws, so the rest of
+# that test method never runs, and *which* tests failed changes the total by
+# itself. Without the failing test names the totals prove nothing.
+#
+# So this instruments before diagnosing. Per run it records:
+#
+#   - the failing test names, not just the counts        (the missing evidence)
+#   - the resolved WP_Sudo autoload root                 (the known confounder)
+#   - whether PHPUnit actually ran, via its exit status  (see below)
+#   - git HEAD and whether the tree was dirty            (was the input identical?)
+#   - PHP version
+#
+# The autoload root is recorded because a symlinked vendor/ makes PHP resolve
+# __DIR__ into another checkout, so the suite runs green — or red — about source
+# you did not change. That happened twice in this project on 2026-07-27 and is
+# the mechanism that reproduces this issue's signature (findings below).
+#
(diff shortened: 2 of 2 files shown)


dknauss:
Closing rather than extending this branch. The current head still has 11 unresolved review findings, including a harness that exits successfully after failed or errored runs, foreign-autoload contamination that can still produce a clean conclusion, invalid-input and output-initialization false greens, incomplete JSON encoding, mutable-HEAD attribution, missing regression coverage, and unregistered PHPUnit behavior claims. Addressing these is a tested redesign rather than a bounded follow-up to the original diagnostic. The reproduced dependency-contamination lesson is retained in AGENTS.md; this executable should not remain open in a misleadingly near-merge state.

Issue #478, opened 2026-07-28
Stale replay-architecture language in code comments (255 in-code matches)


dknauss:
## Problem

4.9.0 removed automatic replay **outright** (#322) — every method, every rule, every admin surface that stashes. The public documentation has now been corrected to say so (#476). The **in-code comments have not**, and they are the larger half of the debt.

Five successive reviews of #476 each found the same defect class one layer further out: stale descriptions of the retired replay architecture, spreading from public docs → internal docs → in-code comments. That pattern is what made #476 unbounded, so the in-code sweep was deliberately re-cut out of it into this issue.

## Why it matters

These are not cosmetic. Reviews of #476 found comments that:

- **Contradict corrected public documentation.** `Gate::evaluate_diagnostic_request()` states *"Since #322 replay **is** authorised by a conjunction this method cannot evaluate,"* naming the binding cookie and `Sec-Fetch-Site` as live conjuncts. `docs/FAQ.md` now says the opposite — there is no resumption for the Rule Tester to predict.
- **Are conditional replay claims** — the `"never replayed unless…"` construction that reads as already-corrected and is not. `Request_Stash::save()`: *"A replay may only run when the confirmation described the WHOLE effect."* `Request_Stash::target_describes_payload()`: *"…such a stash is not eligible for bound replay."*
- **Contradict shipped hook behaviour.** `Event_Recorder::on_action_replayed()`: *"Fired when a stashed request is replayed after successful reauthentication."* `wp_sudo_action_replayed` has no producer in `includes/`, `bridges/`, `mu-plugin/` or `wp-sudo.php` — `readme.txt` correctly calls it retained but dormant.
- **Describe deleted machinery as present.** `admin/js/wp-sudo-challenge.js`'s file header still documents the GET-redirect / POST-self-submitting-form replay engine that the same file's body says was deliberately removed.
- **Document dead data.** `Request_Stash::BINDING_COOKIE` has exactly one reader — `Challenge::clear_binding_cookie()`, a delete. `binding_hash` and `target_complete` are written and read nowhere.

`bridges/` is included because that surface is what third-party integrators read.

## Scope

Inventory-first. Every match is classified **before** any edit:

- **current** — accurate as of shipped behaviour, leave alone
- **stale** — describes the retired design as live, fix
- **historical** — explicitly framed as "an earlier design did X", leave and, where ambiguous, sharpen the framing

[...]

PR #481, opened 2026-07-28
docs(480): stop release-status.md contradicting itself about replay

### LINKED ISSUE #480: docs/release-status.md contradicts itself about replay — in the file CLAUDE.md makes canonical
`CLAUDE.md` names `docs/release-status.md` as the authority **release-state claims MUST be checked against**. It is currently right in three places and wrong in three, about the same fact.

Found during #476's final-SHA confirmation pass. Out of scope for #476 (public boundary correction) and #478 (in-code comments), so tracked here.

## The contradictions

**1. L99** — *"The `#322` v2 layer restored replay and with it the hook, so nothing documented is removed."*

Contradicted by **L129** in the same file (*"the final 4.9.0 payload removes automatic replay outright, so `wp_sudo_action_replayed` has no producer at all"*) and by **L269** (*"the outright removal of automatic replay, which supersedes the v2 origin-bound layer"*).

**2. L148–149** — *"Under v2 it fires only for bound same-browser replays, so it fires **less often** — a narrowed event, not a changed contract."*

Present-tense description of the retired v2 layer, roughly twenty lines below L129–130 saying the same hook has **no producer at all**. The MINOR-not-MAJOR argument is still correct; its supporting sentence describes a layer that was removed.

**3. L100** — the late-freeze bullet describes the retired eligibility architecture as current:
- *"(1) `options.critical` is **now** non-replayable"* — per-rule taxonomy implying unmarked rules replay.
- *"Browser-flow behaviour is unchanged (replay was already refused there); the audit reason moves `incomplete_target` → `replay_blocked`"*.
- *"(4) A new `unnamed_target` reason joins the `wp_sudo_replay_refused` enumeration"*.

**Verified against source:** `Challenge::build_replay_response_data()` assigns exactly two `$refusal_reason` values — `replay_disabled` (L1322) and `no_credential_this_request` (L1333) — and there is a single `do_action( 'wp_sudo_replay_refused', … )` site (L1383). So `unnamed_target`, `incomplete_target` and `replay_blocked` **cannot fire**. The file documents an enumeration that does not exist.

## Correct in the same file, for contrast

L23, L129–136, L165–191, L215–222 and L265–278 all state the removal accurately — L269 even calls the v2 layer superseded and L272 says it is removed. The fix is to bring three passages into line with the rest, not to rewrite the file.

## Acceptance

1. No passage states or implies that automatic replay exists, is conditional, or was restored by the v2 layer.
2. No passage documents a `wp_sudo_replay_refused` reason that cannot fire; where the historical enumeration is worth recording, it is framed explicitly as history.
3. Passages describing the v1/v2 decision remain — they are the versioning rationale — but read unambiguously as history.
4. Docs-only. No behaviour claim changes without being checked against source first.

dknauss:
Closes #480. Found by #476's final-SHA confirmation pass; out of scope for both #476 and #478, so tracked separately.

`CLAUDE.md` names `docs/release-status.md` as the authority release-state claims **must** be checked against. It was right in three places and wrong in three, about the same fact — which is worse than an ordinary stale doc, because a verification rule points at it.

## Three corrections

| Was | Contradicted by | Now |
|---|---|---|
| L99 — *"the #322 v2 layer restored replay and with it the hook, so nothing documented is removed"* | L129, L269 of the same file | Conclusion kept, reason corrected: the hook is **retained but dormant**, not restored |
| L148–149 — *"Under v2 it fires only for bound same-browser replays, so it fires less often"* | L129–130, ~20 lines above | A **darkened** event; MINOR because `wp_sudo_replay_refused` succeeds it at the same lifecycle point |
| L100 — *"`options.critical` is **now** non-replayable"*, *"the audit reason moves `incomplete_target` → `replay_blocked`"*, *"a new `unnamed_target` reason joins the enumeration"* | the source | Marked unreachable, with the history kept |

## Verified against source, not reasoned about

`Challenge::build_replay_response_data()` assigns exactly **two** `$refusal_reason` values — `replay_disabled` and `no_credential_this_request` — from a **single** `do_action( 'wp_sudo_replay_refused', … )` site. So `incomplete_target`, `replay_blocked` and `unnamed_target` cannot fire. The file documented an enumeration that does not exist.

Each is marked unreachable rather than deleted: *why* those reasons were added is the useful part of the record.

## One thing deliberately not called dead

`Action_Registry::stash_no_replay()` is **still live**. `Request_Stash::get_stash_policy()` reads `post_mode => 'none'`, and it genuinely controls whether the POST body is **stored**. The replay framing around it is corrected; the mechanism is described accurately rather than retired by association. Stripping it would have been the easy mistake.

## Untouched

L23, L129–136, L165–191, L215–222 and L265–278 were already accurate — L269 calls the v2 layer superseded and L272 says it is removed. This aligns three passages with the rest of the file rather than rewriting it.

Docs-only.


dknauss:
## Resolution — head `4dbae6f` (rebased onto merged #476)

Both threads asked the same thing: the absolute "never fires" is false because `bin/demo-events.php:179` calls `do_action( 'wp_sudo_action_replayed', … )` when the WP-CLI demo seeder is asked for that event, so subscribers do run. Correct — the source check proves only that `Challenge::build_replay_response_data()` no longer produces it.

**Three sites in this file carried the absolute, not the one cited.** Swept the whole file rather than fixing the flagged line:

| Line | Was | Now |
|---|---|---|
| 99 | "so the hook never fires" | no shipped path fires it |
| 130 | "has **no producer at all**" | no producer on any shipped path, naming the demo seeder |
| 135 | "it simply never runs" | on a shipped path it simply never runs |

The MINOR classification is unaffected: the hook is retained, and its name, signature, documentation, event code `1900009` and subscribers all remain, with `wp_sudo_replay_refused` succeeding it at the same lifecycle point.

Same defect and same fix already applied to `class-event-recorder.php` and the WSAL sensor in #479 — this file was missed because it sits outside that issue's scope.

Rebased onto merged `main` (`3c9b642`); the delta is one file, 13 insertions, and no absolutes survive.

PR #484, opened 2026-07-28
docs(483): five internal docs still described the removed replay engine

### LINKED ISSUE #483: Five internal docs still describe the removed replay engine — two carry executable instructions that would fail
## The gap

4.9.0 removed automatic replay outright (#322). #476 corrected the **public** documentation. #478/#479 corrected the **in-code** comments — that issue is scoped to `includes/`, `admin/js/`, `bridges/`, `mu-plugin/`.

Five files under `docs/` sit between those two scopes and were picked up by neither:

- `docs/ui-ux-testing-prompts.md`
- `docs/security-manual-test-checklist.md`
- `docs/studio-sqlite-release-runbook.md`
- `docs/abilities-api-assessment.md`
- `docs/core-sudo-gate-implementation-spec.md`

Found by the final-SHA confirmation pass on #476, which swept outside the PR's own files.

## Why two of these are worse than stale prose

`AGENTS.md` grades prose that says what to run like code. Two of these tell a human to do something and expect a result that cannot happen:

- **`docs/studio-sqlite-release-runbook.md`**, "Release Verification Steps" step 4: *"replay a gated settings POST after auth"*, and a sibling step expecting a challenge URL's `return_url` to be honoured. `return_url` is inert since 4.9.0 and `Challenge::build_replay_response_data()` refuses to consult it even to classify the landing. **A release engineer following this reports a false regression.**
- **`docs/security-manual-test-checklist.md`**, §1 "Non-replayable stash": *"confirm the original save applied once"* — the challenge applies nothing; the user re-issues the action.

`docs/ui-ux-testing-prompts.md` is the same class: six sites instruct a tester to verify a `"Replaying your action…"` overlay and a `wp.a11y.speak()` replay announcement. That string does not ship — the localized string is `'returningToPage' => "Returning you to your page…"`, and `handleReplay()` assigns it unconditionally.

`docs/abilities-api-assessment.md` and `docs/core-sudo-gate-implementation-spec.md` are prose-only: present-tense "stash-and-replay" descriptions, the latter contradicting its own §5.2/§11.

## How it happened — worth recording

#476 was originally one large change. After five review rounds it was re-cut into a bounded public-boundary PR (subset A) plus a deferred remainder (subset B). Subset B was assumed to be covered by the in-code sweep, but that sweep's scope is code paths, not `docs/`. **The re-cut created a seam and nothing owned it.**

The corrected versions of all five already exist — they were written during #476's round 3 and set aside with subset B. They are being recovered rather than rewritten.

## Scope

Docs-only. No behaviour claim changes without being checked against source first. Explicitly **not** in scope: the in-code comments (#478/#479), `docs/release-status.md` (#481), or the `is_handler_endpoint()` execution-context fix (separate PR).

## Acceptance

1. No instruction tells a human to expect a replay, a resumed action, or a `"Replaying your action…"` overlay.
2. No document states or implies automatic replay exists, is conditional, or can be restored.
3. Prose describing the retired design as history is explicitly framed as history.
4. Executable steps are verified against the code they exercise, not just reworded.

dknauss:
Closes #483. Recovered from work set aside during #476's re-cut.

## The seam

4.9.0 removed automatic replay outright (#322). #476 fixed the **public** docs. #478/#479 fixed the **in-code** comments — scoped to `includes/`, `admin/js/`, `bridges/`, `mu-plugin/`. These five live under `docs/` and were covered by neither.

That gap is a direct consequence of re-cutting #476: subset B was set aside on the assumption the in-code sweep would take it, and the in-code sweep is about code paths. Nothing owned the remainder. It was found by the **final-SHA confirmation pass on #476**, which swept outside the PR's own files — worth noting, because no reviewer scoped to a PR's diff would have seen it.

## Two are executable instructions that would fail

`AGENTS.md` grades prose that says what to run like code.

| File | Said | Reality |
|---|---|---|
| `studio-sqlite-release-runbook.md` | "replay a gated settings POST after auth"; expect `return_url` to be honoured | `return_url` is inert since 4.9.0; `build_replay_response_data()` refuses to consult it even to classify the landing. **A release engineer following this reports a false regression.** |
| `security-manual-test-checklist.md` | "confirm the original save applied **once**" | The challenge applies nothing; the user re-issues. |

## One correction inside the correction

The runbook step now says where a gated settings POST *actually* lands: the **dashboard**, not Settings — `options.php` is in `Challenge::HANDLER_ENDPOINTS` because a bare GET of it renders the raw All Settings dump.

An earlier draft of this very fix said "the originating screen." True for ordinary admin URLs, false for exactly the case the step exercises. Caught by applying a finding from an open #476 review thread before this reached a reviewer.

## The other three

- **`ui-ux-testing-prompts.md`** — six sites instructing testers to verify a `"Replaying your action…"` overlay and a `wp.a11y.speak()` replay announcement. That string does not ship; `handleReplay()` assigns `'returningToPage'` ("Returning you to your page…") unconditionally.
- **`abilities-api-assessment.md`** — two present-tense "stash-and-replay" descriptions.
- **`core-sudo-gate-implementation-spec.md`** — a stashed-POST-replay reference contradicting its own §5.2/§11.

## Recovered, not rewritten

These corrections were written during #476's round 3 and set aside with subset B. Extracted **path-by-path** from the stash rather than popped — the same stash also holds twice-rejected versions of `class-challenge.php`, `class-request-stash.php`, `wp-sudo-challenge.js` and `current-metrics.md`, and popping it would have overwritten landed corrections under a valid approval. Verified: those six files are byte-identical to `origin/main` in this branch.

Docs-only.


dknauss:
Triage of the eight open Codex threads, verified against the current head (`b41f7a9`) rather than against the repair commit messages. All eight are addressed; resolving them.

| Finding | Where it is now satisfied |
|---|---|
| Separate direct re-POST from stash-key reuse | `security-manual-test-checklist.md` L91-99 now carries two separate checks — "Consumed stash key is dead" and "A user's own re-POST still works, and that is correct", the latter stating the session remains active and `stash_no_replay()` governs automatic replay, not deliberate duplicate requests |
| Describe the remaining uses of the request stash | `abilities-api-assessment.md` L259-261 now names the URL choosing the landing, the flag selecting the notice, and the rule ID for `wp_sudo_replay_refused`, and says it is not dead data |
| Cite the registered `options.php` source | `studio-sqlite-release-runbook.md` L67 cites `GB-OPTIONS-ALLSETTINGS` |
| Navigate explicitly back to the settings screen | Same file L67-68: "navigate **explicitly** to Settings and re-submit — browser Back returns to the challenge page, not the form" |
| Expect the resume page for a consumed stash | `ui-ux-testing-prompts.md` L121-123 and `security-manual-test-checklist.md` L93 both now expect "Session already confirmed" |
| Expect the dashboard for a direct `return_url` challenge | Runbook L69-73 splits the mode out and expects the dashboard, and calls navigation to the supplied destination a regression |
| Clear the sudo session before the gated POST | Runbook L62: the gated POST bullet now opens "with no active sudo session", after the acquire/deactivate bullet |
| Scope the return announcement to stashed challenges | `ui-ux-testing-prompts.md` L20 is scoped "(stashed challenge)" and states session-only redirects directly without the intermediate announcement |

One instance of the same defect was left standing outside the threads, fixed in `b41f7a9`: the H9 checklist claimed the "Your challenge session has expired" message appears whenever the stash is consumed or timed out. That holds only with no sudo session active — `render_page()` enters `render_resume_page()` before checking whether the stash exists, so with a live session the same URL shows "Session already confirmed". It contradicted the correction made two sections below it.

PR #512, opened 2026-07-29
fix(deps): update WPCS past CVE-2026-45293

### LINKED ISSUE #511: Update WPCS past CVE-2026-45293
## Problem

The Composer security audit now reports high-severity advisory GHSA-3pwp-g2mj-5p3v / CVE-2026-45293 against the locked `wp-coding-standards/wpcs` 3.3.0 development dependency. The affected range is `>=0.14.1,<3.4.1`. This currently fails the Composer Audit check on otherwise unrelated pull requests.

## Acceptance criteria

- Update the lock to WPCS 3.4.1 or later without broad unrelated dependency churn.
- `composer audit --locked --no-interaction` passes.
- Unit tests, lint, and static analysis pass.
- Confirm the worktree uses an isolated, non-symlinked `vendor/`.

Source: https://github.com/advisories/GHSA-3pwp-g2mj-5p3v

dknauss:
Closes #511.

Updates the locked WordPress Coding Standards dependency from 3.3.0 to 3.4.1, the first version outside the affected range for GHSA-3pwp-g2mj-5p3v / CVE-2026-45293. Composer also advances three tightly coupled PHPCS support packages required by the patched release.

Validation:
- isolated non-symlinked worktree vendor and autoload-root check
- `composer audit --locked --no-interaction`
- `composer test` (1,301 tests, 3,893 assertions)
- `composer lint`
- `composer analyse`
- `composer verify:metrics`

Advisory: https://github.com/advisories/GHSA-3pwp-g2mj-5p3v

### DIFF
--- a/.sbom/bom.json
+++ b/.sbom/bom.json
@@ -98,11 +98,11 @@
             "properties": [
                 {
                     "name": "cdx:composer:package:distReference",
-                    "value": "c58ee7d5d20eb86f409b0b7f65d47f97470114e9"
+                    "value": "8c1ac274ae81fb1601a1d6098a4b08295801eb6d"
                 },
                 {
                     "name": "cdx:composer:package:sourceReference",
-                    "value": "c58ee7d5d20eb86f409b0b7f65d47f97470114e9"
+                    "value": "8c1ac274ae81fb1601a1d6098a4b08295801eb6d"
                 },
                 {
                     "name": "cdx:composer:package:type",
@@ -365,10 +365,10 @@
             ]
         },
         {
-            "bom-ref": "amphp/parallel-2.3.3.0",
+            "bom-ref": "amphp/parallel-2.4.0.0",
             "type": "library",
             "name": "parallel",
-            "version": "v2.3.3",
+            "version": "v2.4.0",
             "group": "amphp",
             "description": "Parallel processing component for Amp.",
             "author": "Aaron Piotrowski, Niklas Keller, Stephen Coakley",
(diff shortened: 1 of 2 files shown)


dknauss:
Re: the SBOM finding (the inline thread's anchor went stale after the push, so replying here).

Correct, and material — fixing the lock without the SBOM leaves the artifact consumers actually read still advertising the CVE. Regenerated in `e3ab571`.

`wp-coding-standards/wpcs` is now 3.4.1, with `phpcsextra` 1.5.1, `phpcsutils` 1.2.3 and `phpcodesniffer-composer-installer` v1.2.1. No `3.3.0` string and none of the old git refs survive anywhere in the file — not in purls, bom-refs or hashes. Every name/version pair was compared programmatically against `composer.lock`: 100 packages, zero mismatches, zero missing in either direction. Byte-identical on a second `composer sbom` run, so it is generated rather than hand-patched.

**Heads-up on the diff size**, since it looks alarming: 271 of 541 component entries move, not the four packages this PR names. That is pre-existing drift rather than scope creep — the committed SBOM was already stale before this branch (`phpstan/phpstan` listed at 2.1.42 while `composer.lock` moved to 2.2.6 in `93bf57c`, plus several `amphp/*` transitives). A full resync against the lock is what `composer sbom` produces and what the file is meant to contain.

That staleness is its own finding: nothing regenerates this file when dependencies change, so it drifts silently until someone looks — which is exactly how a shipped SBOM ends up advertising a CVE the lock had already fixed. Filed as a follow-up rather than widening this PR.

Issue #530, opened 2026-07-29
Integration test: assert a second administrator gets no governance caps on fresh activation


dknauss:
Follow-up from #524. The unit test that ships with that fix asserts **call ordering**, which is a proxy for the security property rather than the property itself.

## What is covered now

`PluginTest::test_activate_grants_activating_admin_before_running_the_upgrader` records a call sequence and asserts the activating admin holds `manage_wp_sudo` **before** `upgrade_3_3_0()`'s zero-holder check runs. It was falsified against the pre-fix commit — a scratch worktree at the unfixed ordering fails it with `Failed asserting that 1 is less than 0` — so it is a real regression guard, not a tautology.

It is asserted that way because `add_cap()` on a mocked `WP_User` cannot change what a stubbed `get_users()` returns. At unit level, ordering is the observable behaviour.

## What is not covered

**The actual property.** Nothing asserts that a *second, non-activating* administrator ends up without `Admin::GOVERNANCE_CAPS`.

That gap has a concrete failure mode: if someone changed `upgrade_3_3_0()`'s holder check from a `capability` query to a role-based one, the broad grant would return **with the ordering intact** — and the shipped unit test would still pass.

This is also why the bug survived in the first place. `tests/Integration/UpgraderTest.php:398` (`test_first_activation_grants_manage_wp_sudo_to_administrator`) asserts only that the *activating* admin received the capabilities, which was true both before and after the fix.

## Wanted

An integration test under `tests/Integration/`:

1. Two administrator accounts present **before** activation.
2. Run the activation path as user A.
3. Assert user A holds all four `Admin::GOVERNANCE_CAPS`.
4. Assert user B holds **none** of them.

Worth adding a companion for the no-current-user path (`get_current_user_id() === 0`), where the backfill legitimately *does* grant broadly, so the test pins both branches rather than only the one being protected.

## Why it was not shipped with #524

Neither the implementing session nor the reviewer could execute the integration suite in that worktree — no `.tmp/wordpress-tests-lib` present, and installing one risks a test database other concurrent sessions are using. A test that neither party has watched fail is worse than an absent one: it reads as coverage in the file listing and in `docs/current-metrics.md` while asserting nothing anyone has verified.

Filed rather than shipped unexecuted, deliberately.

## Corroboration that the fixed behaviour is the specified one

[...]

Issue #404, opened 2026-07-27
Upgrade replay from 0.0.0 silently grants governance caps to every administrator


dknauss:
Surfaced by the Coordination session while removing a bad `wp_sudo_db_version` remedy from #398 (`7579c4d`, llm-lies #55). **Removing the command closes the advertised route, not the underlying fail-open** — so filing it.

## Verified against `origin/main`

- `Upgrader::get_db_version()` reads `get_site_option()` on multisite and `get_option()` otherwise, so the stamp is one network-wide row on multisite.
- `upgrade_2_0_0()` calls `remove_role( 'site_manager' )` unconditionally.
- `upgrade_3_3_0()` (single-site only) queries for one holder of `manage_wp_sudo`; **if there are none it adds all four `GOVERNANCE_CAPS` to every administrator.**

## The fail-open

The routine cannot distinguish *"this install predates governance caps"* from *"this install deliberately has none."* Both look like zero holders.

For a genuine pre-3.3.0 upgrade, granting to admins is the correct bootstrap. For an install that **deliberately revoked governance** from all administrators and relies on scoped `WP_SUDO_RECOVERY_MODE`, a replay from `0.0.0` hands governance back to every administrator — silently, with no audit event, in a security plugin whose stated model is that governance is separate from `manage_options`.

## Reachability without the removed command

The stamp can be absent for reasons other than following bad documentation: a restore or migration that drops the option row, a staging→production copy, or any operator who clears it. The consequence does not depend on *how* it went missing.

## Possible dispositions

1. **Distinguish replay from bootstrap.** A missing stamp *plus* existing plugin state (e.g. `wp_sudo_settings`) is anomalous — a true fresh install has neither, a true legacy upgrade has a non-zero stamp. Skip the grant in that case and surface it instead.
2. **Never grant silently.** Fire an audit action (or a Site Health critical) whenever the routine grants caps, so the escalation is at least visible.
3. **Accept and document** in `security-model.md`, as a known consequence of clearing the stamp.

(1)+(2) is my preference; (3) alone leaves a silent privilege grant reachable by accident.

## Not fixed by #393

#393's observation — that the `>=` guard means the defect expires once no future routine is keyed at or below the stamped version — addresses the **stamp value**. It does not address a replay from `0.0.0`, which runs `upgrade_3_3_0()` regardless.

dknauss:
[...]

PR #531, opened 2026-07-29
fix(524): grant the activating admin before running the upgrader

### LINKED ISSUE #524: Fresh activation grants governance caps to EVERY administrator, not just the activating one
Found during the #404 design review, verified directly against `main` at `46718c5`. Probably worse than #404 itself, because it fires on **every fresh install** rather than only after a lost version stamp.

## The ordering

`Plugin::activate()` (`includes/class-plugin.php:622`):

```php
public function activate(): void {
    // Run the upgrader to stamp the version on fresh installs.
    $upgrader = new Upgrader();
    $upgrader->maybe_upgrade();          // <-- line 624-625
    ...
    // Grant governance capabilities to the activating admin ...
    $admin = get_userdata( get_current_user_id() );
    if ( $admin instanceof \WP_User ) {
        $admin->add_cap( 'manage_wp_sudo' );   // <-- line 637-643
        ...
    }
    update_option( 'wp_sudo_activated', true );
}
```

On a fresh single-site activation the stored version is `0.0.0`, so `maybe_upgrade()` runs every routine — including `upgrade_3_3_0()`, which queries for one holder of `manage_wp_sudo`, finds **none** (nothing has granted it yet), and therefore adds all four `Admin::GOVERNANCE_CAPS` to **every user in the administrator role**.

Only afterwards does line 639 grant the same four caps to the activating admin — redundantly, since the loop already covered them.

## Why it matters

The plugin's stated model is that governance is deliberately **separate** from `manage_options`. `upgrade_3_3_0()`'s own docblock describes granting "to the activating admin". The ordering defeats both: on a multi-administrator site, activating WP Sudo hands governance to every administrator at once, silently.

This is the same effect #404 describes, reached by the ordinary install path rather than by losing the version stamp.

## Why no test caught it

`test_first_activation_grants_manage_wp_sudo_to_administrator` (`tests/Integration/UpgraderTest.php:398`) asserts only that **the activating admin** ends up with the caps. That is true either way, so the test passes while the behaviour is wrong. Nothing asserts that a *second, non-activating* administrator does **not** receive them.

## Suggested fix

Move the activating-admin grant **above** `maybe_upgrade()`, so the zero-holder check in `upgrade_3_3_0()` sees a holder and correctly skips the broad bootstrap on fresh installs.

That ordering also composes correctly with the #404 guard: on a fresh activation `wp_sudo_activated` is not yet written (it is the last line of `activate()`), so a state-based replay guard does not fire and legitimate bootstrap is unaffected.

Wants a regression test asserting a second administrator does **not** receive `GOVERNANCE_CAPS` on fresh activation.

## Caveats to check when fixing

- `Plugin::activate_network()` (`class-plugin.php:663`) also calls `maybe_upgrade()`; `upgrade_3_3_0()` returns early on multisite, so network activation is unaffected — confirm rather than assume.
- Under WP-CLI `get_current_user_id()` is `0`, so `get_userdata()` returns `false` and the grant is skipped; moving it earlier must not change that, and must not leave a CLI-activated site with zero holders *and* a skipped bootstrap.

## Related

- #404 — the replay case; this is the fresh-install case of the same effect

dknauss:
Closes #524.

On **every fresh single-site install with more than one administrator**, activating WP Sudo granted the four governance capabilities to *all* of them at once — silently, in a plugin whose stated model is that governance is deliberately separate from `manage_options`.

## The ordering

`Plugin::activate()` called `maybe_upgrade()` before granting caps to the activating admin:

```php
$upgrader = new Upgrader();
$upgrader->maybe_upgrade();          // <- ran first
...
$admin = get_userdata( get_current_user_id() );
if ( $admin instanceof \WP_User ) {
    $admin->add_cap( 'manage_wp_sudo' );   // <- four statements later
```

A fresh install has no stored version, so every migration runs — including `Upgrader::upgrade_3_3_0()`, which queries for one holder of `manage_wp_sudo` and, finding **none**, adds all four `Admin::GOVERNANCE_CAPS` to every user in the administrator role. None existed yet, because the grant that would have created one came after.

`docs/security-model.md:53` has always said *"the activating admin receives all four capabilities. Other admins receive none until explicitly granted."* **That sentence was false before this commit and is true after it.**

## The fix

Move the activating-admin grant above `maybe_upgrade()`, so the zero-holder check finds a holder and skips the broad bootstrap — which is what `upgrade_3_3_0()`'s own docblock always described.

The backfill keeps the role it was written for: an activation with **no current user** — `wp plugin activate` without `--user` — where the direct grant cannot happen and a site with no governance holder at all would be worse. With `--user`, the narrow grant happens exactly as in a browser activation.

## Verification

The test asserts **call order**, because `add_cap()` on a mocked `WP_User` cannot change what a stubbed `get_users()` returns — the ordering is the observable behaviour at unit level.

The reviewer independently falsified it rather than reasoning about it: a throwaway worktree at the pre-fix commit with this test dropped in **fails** (`Failed asserting that 1 is less than 0`, sequence `['zero_holder_check','activating_admin_granted']`). It passes on this branch both isolated and in the full suite.

Also verified by the reviewer against real WordPress source: a `capability` user query builds `{prefix}capabilities LIKE '%"manage_wp_sudo"%'`, so a capability added directly by `add_cap()` **does** match the zero-holder check; and `default-filters.php` hooks `wp_cache_set_users_last_changed` to `updated_user_meta`, so the users-query cache cannot serve a stale pre-grant result. The fix works for the reason claimed, not incidentally.

`activate_network()` is unaffected — `upgrade_3_3_0()` early-returns on multisite (verified, not assumed). The relative order of `maybe_upgrade()` → `strip_editor_unfiltered_html()` → `schedule_prune_cron()` is unchanged; only the grant moved.

## Operator impact

**Existing installs: nothing changes.** The version stamp is already current, so the upgrader returns early and no re-grant was occurring. This affects fresh activations only.

No capabilities are revoked. An install set up while several administrators existed should review Settings → Sudo → Access.

## Documentation swept

Found by the reviewer sweeping file types rather than the diff:

- `docs/developer-reference.md` — "First-run lockout safety" described the backfill as the mechanism providing the guarantee. It now describes the direct grant, says the ordering is load-bearing and why, and keeps the backfill's real remaining role.
- `docs/release-status.md` — still claimed post-`v4.9.2` drift was "one commit … No payload drift". This commit is payload.
- `CHANGELOG.md` — no entry existed for a user-visible security change.

## Known limitation, tracked separately

The unit test locks the **ordering**, which is a proxy for the security property. It would not catch someone changing `upgrade_3_3_0()`'s holder check to a role-based query, which would reinstate the broad grant with the ordering intact.

The real assertion needs integration: two administrators, activate, assert admin #2 holds none of `GOVERNANCE_CAPS`. Filed as **#530** with the exact property and both branches to pin. Not added here because neither I nor the reviewer could run the integration suite in this worktree — no `.tmp/wordpress-tests-lib`, and installing one risks a test DB other concurrent sessions may be using. Shipping a test neither of us has executed would be worse: it reads as coverage in the file listing and in `docs/current-metrics.md` while asserting nothing anyone has watched fail.

Corroboration that the fixed behaviour is the specified one, found by the reviewer sweeping file types: `tests/MANUAL-TESTING.md` §23 already tells the tester the second administrator "does **not** yet hold `manage_wp_sudo`". If both accounts existed at activation, that was unachievable as written before this fix — the manual checklist has been describing intended behaviour the code did not deliver.

`tests/Integration/UpgraderTest.php:398` still passes, and always would have: it asserts only that the *activating* admin received the capabilities, which was true either way. That is exactly why the bug survived.

## Origin

Found during the **design review for #404**, which covers the related case — a *replayed* migration after the version stamp is lost. That review's verdict is recorded on #404 and also corrected the discriminator I had planned to use there.

## Gates

`composer test:unit` 1308 tests / 3907 assertions · `lint` 24/24 · PHPStan level 6 no errors · Psalm no errors · `verify:metrics` in sync (re-measured) · `verify:i18n` in sync · `bin/make-pot.sh` idempotent.


dknauss:
One prose correction is needed before merge: `UpgraderTest::test_first_activation_grants_governance_cap()` still says twice that `Plugin::activate()` runs the upgrader *then* grants caps (docblock around line 393 and Act comment around line 414). This PR deliberately reverses that order, so those unchanged comments now contradict the tested behavior. Please update them to say the activating admin receives the direct grant before `Upgrader::maybe_upgrade()`; the stacked #530 integration test then proves why that order matters.

dknauss:
All five findings addressed in `5ff112a` — four applied, one declined with reasoning. Replying here rather than in-thread since the pushes have outdated several anchors.

**P1 — impact statement.** You were right, and the first correction was wrong in the *opposite* direction, which is worth recording. It initially said "existing installs: nothing changes / fresh activations only", missing the deactivate → upgrade → reactivate cycle. The fix for that then claimed the cycle re-grants generally — false for essentially the whole install base, because `upgrade_3_3_0()` is keyed `3.3.0` and the loop guard is `version_compare( $stored, '3.3.0', '<' )`. A site stamped 4.9.x never reaches the routine. It now scopes to **below 3.3.0 or missing**, names the guard, and conditions the actionable line on the stamp itself rather than on the generalisation above it.

**P1 — integration-test comments.** Corrected in both places. Also added why the test did not catch this: it asserts only that the *activating* admin received the caps, which was true under both orderings. The missing assertion — a second administrator receives none — is #530.

**P1 — WP-CLI claim registered** as `GB-CLI-NO-USER`, anchored on `wp_set_current_user( $user->ID );` in `Runner.php` (verified unique). One correction to my own first draft, caught in review: the gate is a configured `user` **value**, not the flag — `config-spec.php` gives `user` a `file` entry, so a `user:` key in `wp-cli.yml` populates it with no flag present, and such a site *does* establish a current user. The row says the scope the anchor proves rather than an absence claim over all of wp-cli.

**P2 — `finally` block** added, matching the pattern already used twice in that file.

**P2 — rollback on aborted migration: declining.** `Plugin::deactivate()` restores `unfiltered_html` and clears crons but **never removes governance capabilities**, so caps surviving a clean deactivate/reactivate is deliberate success-path behaviour. An error-path rollback would make the *failed* activation stricter than the successful one.

Worth noting I abandoned my first reason for declining. I argued a rollback would need new state to distinguish "we granted this" from "they already had it" — that is wrong, a local `has_cap()` capture before granting suffices. The argument above is the one the evidence supports. The realistic failure modes are also an uncatchable fatal (a `catch` never runs) or a silent `dbDelta` no-op (nothing to catch), and the residual is one extra administrator who deliberately attempted activation. Happy to be pushed back on.

**Known, not fixed:** "never reaches the routine" has a narrow false case — `set_db_version()` runs *after* the loop by design, so a site fatalling mid-migration never gets stamped and stays in the affected cohort. The stamp-conditioned instruction still routes it correctly, which is why this is a clause for the next touch rather than a restage.

Issue #537, opened 2026-07-29
Pre-gate protected controls in place: disable, explain, reauthenticate, unlock — implementation of #436 face 4


dknauss:
Focused implementation issue for **#436 face 4**, from a design direction given by the repo owner. Face 4 was already identified in the #436 design review as the most valuable of the remaining faces; this scopes it, because it spans several surfaces and needs a per-control inventory.

## The principle

For discrete dangerous controls, **prevention before input beats accepting the action, discarding it, and explaining afterward.**

The pattern:

- Leave ordinary editing available.
- Disable **only** gated controls: role changes, password/email changes, delete/remove actions, bulk role changes, and similar privileged operations.
- Show why they are unavailable — "Reauthenticate to change protected fields."
- Provide a nearby reauthentication affordance.
- After successful reauthentication, **unlock the controls in place**. Never execute the original action automatically.
- Keep the server-side gate authoritative. **Disabled controls are UX, not a security boundary.**

This is how the pre-4.9 feeling is recovered without restoring the automatic replay #322 removed for cause.

## Do NOT disable

- The entire profile form.
- The general "Edit user" link.
- Navigation to user pages.
- **Any control whose omission from the POST might itself change data**, until WordPress's handler behaviour has been verified for that specific field.

### That last one is not hypothetical — this codebase has already been bitten by it

A disabled input is **not submitted**. WordPress's Settings API treats an absent allowlisted option as a write, not a no-op:

- `wp-admin/options.php` initialises `$value = null` for every allowlisted option and calls `update_option()` unconditionally (`GB-OPTIONS-NULL-WRITE`).
- `sanitize_option()` turns a null `default_role` into `subscriber` (`GB-DEFAULT-ROLE-NULL`) — a *less restrictive* value.
- `absint( null )` flips `users_can_register` from 1 to 0.

That is #445, and #506 is the same mechanism reached through the unregistered-settings page. **So on Settings screens, disabling a checkbox can be the write.** Each candidate field needs its handler behaviour checked before it is disabled, not after.

## Per-control inventory required

| Surface | Controls | Notes |
|---|---|---|
| **User profile** (`profile.php`, `user-edit.php`) | role, password, email, application passwords | Highest value. Lets ordinary fields — display name, biography, preferences — be edited safely without a mixed form that loses everything on submit |
| **Users list** (`users.php`) | delete / remove, bulk delete, bulk role change | |
[...]

PR #540, opened 2026-07-29
fix(533): land a refused user-edit POST on the Users list, not the dashboard

### LINKED ISSUE #533: Refused user-edit.php POST lands on the dashboard because query-stripping removes user_id — not the same defect as #436's faces
Reported from live use: changing another user's information or role without a sudo session sends you to the **dashboard** after reauthenticating, not back to the user you were editing. The typed input is gone either way (that is #436 face 1), but losing your *place* is a separate defect with a separate cause, and it is the cheapest of the three to fix.

## The cause chain

`Challenge::build_replay_response_data()` derives the landing from the server-recorded `$stash['url']`, then **strips the query** — because for a gated request the query carries `action` + `_wpnonce`, which *is* the effect. That is correct and must stay.

But stripping the query off `user-edit.php?user_id=5` yields a bare `user-edit.php`, and core's `user-edit.php` calls `wp_die( 'Invalid user ID.' )` when `$user_id` is 0 and it is not the profile page (`GB-USER-EDIT-DIES` in `docs/upstream-sources.md`). So the file is listed in `Challenge::HANDLER_ENDPOINTS`, and handlers are routed to the neutral page rather than stranding the user on a `wp_die()`.

**Every step is individually correct. The dashboard landing is an emergent consequence of query-stripping, not a property of the screen.**

## The asymmetry that shows it

- Editing **your own** profile → lands back on `profile.php`. It is not in `HANDLER_ENDPOINTS`, because it defines `IS_PROFILE_PAGE` and renders fine with no query.
- Editing **another user** → dashboard.

Same rule (`user.promote_profile` / `user.change_password` / `user.change_email` all match both `pagenow` values), same lost input, entirely different landing — decided by whether the target screen happens to need a query parameter to identify itself.

## Why this is not #436 face 1, 3 or 4

- **Face 1** is preserving the typed input. Fixing this does not restore a single field.
- **Face 3** is naming every rule a form's changes matched.
- **Face 4** is gating before the form is offered.

This is a fourth thing: **landing on the correct screen at all.** It has been folded into "the landing is bad" and so has never been scoped on its own. It is also the one that reads as a plain bug rather than a security tradeoff — nobody chose to send user-edit to the dashboard; it fell out.

## Shape of a fix, and the trap in it

Preserve **identifying, non-effect** query parameters for known targets — `user_id` for `user-edit.php` — while continuing to strip everything else.

The obvious naive version is dangerous and should be rejected explicitly: *do not* pass the query through, and *do not* allowlist by "params that look harmless". A third-party rule can gate an effect that fires on a bare-path GET with a query parameter, which is precisely the confused deputy #322 removed. The allowlist must be **per known screen**, small, and justified per entry.

Worth stating for whoever picks this up: `$stash['url']` is the URL the gate intercepted, not the requester-supplied `return_url` — so this does not reintroduce requester control over the destination the way honouring `return_url` would. But the intercepted URL is still attacker-*influenceable* in the confused-deputy scenario, which is why the answer is an allowlist of parameters that cannot carry an effect on GET, not a general relaxation.

For `user_id` on `user-edit.php` specifically: a GET renders the edit form and performs no state change, so landing there executes nothing. That reasoning needs to be re-derived for any second entry added to the list.

## Requires a design review before code

Per `CLAUDE.md` — this touches the landing path, which is security-sensitive, and the guard is exactly the kind of allowlist that is wrong when it is too wide *and* useless when it is too narrow.

Specific things a reviewer should be asked to break:
- whether `user_id` can reach any effect on a GET of `user-edit.php` under any built-in or plausible third-party rule
- whether the same treatment is safe for network `user-edit.php` (`is_handler_endpoint()` decides from the URL path, deliberately not from `is_network_admin()`)
- whether preserving a parameter changes what `Gate::match_request()` sees on the landing GET, since the landing re-enters the gate with a now-active session

## Related

- #436 — faces 1, 3, 4; this is the fourth, distinct
- #429 — established that a refused POST should return to its originating screen; this is the case where that intent is defeated by the query strip
- #322 — why the query is stripped in the first place

dknauss:
Closes #533. Reported from live use: changing another user's information or role without a sudo session cost you your typed input **and** your place — you came back on the dashboard.

## What was happening

`user-edit.php` is in `Challenge::HANDLER_ENDPOINTS` because a bare GET of it calls `wp_die( 'Invalid user ID.' )` (`GB-USER-EDIT-DIES`). Handlers were diverted to the neutral page, because the dashboard was the only alternative to stranding someone on a dead end.

That classification is correct and unchanged. What changes is that a handler with an **unambiguous usable sibling** now lands there instead. `user-edit.php` has one: the Users list.

## My first proposal was wrong, and the design review caught it

I filed #533 claiming the cause was query-stripping discarding `user_id`. **That was false**, and I'd inferred it rather than verified it.

Core's profile form posts to a queryless `self_admin_url()`, with `action=update` and `user_id` as **hidden POST fields** (`user-edit.php` L277, L959-960, read from `$_REQUEST` at L15-16). The stashed URL carries no query at all, so the strip never fires and a parameter allowlist would have been a **no-op**.

Worse: `tests/Unit/ChallengeTest.php` carried a fixture `user-edit.php?user_id=42` — the only query-bearing `user-edit.php` stash URL in the suite, and **no live request produces one**. A TDD cycle seeded from that shape yields a green test for a guard that never fires. It is now the shape a real request makes, with a comment saying why.

The issue body is [corrected in-thread](https://github.com/dknauss/Sudo/issues/533).

## The mechanism

A separate `HANDLER_LANDINGS` map, keyed on the handler's filename. The destination is **static**: nothing in the request body and no requester-supplied `return_url` can name it — that bound is what distinguishes this from honouring `return_url`.

The request path selects only between the site and network forms of that same target, **and the capability checked with it**:

| Screen | Capability |
|---|---|
| `wp-admin/users.php` | `list_users` |
| `wp-admin/network/users.php` | `manage_network_users` |

A single capability would have sent a multisite administrator who is **not** a super admin from a working dashboard to a 403 — the exact outcome the check exists to prevent. Found in review; now has its own regression test.

Network context comes from the URL path, not `is_network_admin()`, matching `is_handler_endpoint()` — `GB-IS-NETWORK-ADMIN` records why that function can't classify a URL.

The map is deliberately **partial**. `options.php` gets no entry: `options-general.php` is one of several Settings screens, and guessing is worse than the dashboard.

## The guard test earned its place immediately

`test_every_handler_landing_target_is_a_verified_screen()` walks the map in both site and network form against the suite's verified-screen set, and asserts each entry declares all three keys.

**It failed on its first run** — network `users.php` was absent from the verified set, which is the same gap that made the wrong capability invisible. Adding a map entry now forces someone to decide what that screen does on a bare GET.

What it does **not** establish: which capability each target requires. `verifiedScreens()` holds URLs. That half is recorded against `GB-USERS-LIST-CAP` and `GB-NETWORK-USERS-BARE-GET` — both machine-rechecked by `verify:sources` — and remains a reviewed claim rather than a tested one. The CHANGELOG says so explicitly; an earlier draft claimed the test established both, contradicting the test's own docblock.

## Why the docblocks are short

Four review rounds found **one real bug and three defects in prose written to explain it** — all three in justification clauses ("the ambiguity costs nothing", "keeps both decisions on the same footing", "and which capability it needs"). Each asserted a conclusion the surrounding facts didn't carry, and nothing tests a conclusion.

So the docblocks now carry facts and pointers to the checks that enforce them, and stop before the sentence saying what the facts mean. The rejected-design rationale lives on #533.

The reviewer verified the cut changed no behaviour by tokenising `class-challenge.php` at both trees, dropping comments, and comparing: **24797 tokens each side, identical**.

## Split into two commits

The 11-file change exceeded the 10-file limit. `b20d7c4` is the mechanism plus its tests, registry rows and metrics; `5160cf2` is the prose sweep. Mechanism first deliberately — between them the docs under-describe behaviour that exists, rather than promising behaviour that doesn't.

The prose sweep is its own evidence of the cost here: **one behaviour change falsified four copies of one sentence across three files**, plus an unconditional "or the neutral page" and a "regardless" that AGENTS.md specifically regulates.

## Scope

Does **not** restore the typed input (#436 face 1). Not the structural fix either — pre-gating protected controls so the work is never lost is #537. This is the part that was a plain bug rather than a tradeoff.

## Verification

`composer test:unit` 1313 tests / 3931 assertions · `lint` 24/24 · PHPStan level 6 + Psalm clean · `verify:metrics` in sync · `verify:i18n` in sync · `verify:sources` 69 citations present.

The reviewer additionally verified the first commit stands alone by materialising its tree in isolation and running `verify-metrics`, `verify-i18n`, `verify-sources`, PHPStan, Psalm and PHPCS against that copy.


dknauss:
Closing as not planned because the WP Sudo program is concluded. The user-edit landing fix remains documented in this PR, but further product UX work would imply a supported installable product that no longer exists. PR #545 is the canonical conclusion record.

PR #541, opened 2026-07-29
fix(404): refuse to re-grant governance when the version stamp is lost

### LINKED ISSUE #404: Upgrade replay from 0.0.0 silently grants governance caps to every administrator

dknauss:
Closes #404 (security, priority: high).

`Upgrader::upgrade_3_3_0()` grants all four `GOVERNANCE_CAPS` when it finds no holder of `manage_wp_sudo`. It could not distinguish **"this install predates governance caps"** from **"this install's version stamp went missing"** — both present as zero holders.

So a stamp lost to a restore, a staging→production copy, a migration tool, or an operator clearing the row replayed every routine from `0.0.0` and handed governance back to **every administrator** — silently, with no audit trail, in a plugin whose stated model is that governance is deliberately separate from `manage_options`.

## The discriminator

`wp_sudo_activated`. `Plugin::activate()` writes it *after* `maybe_upgrade()`, and `uninstall.php` deletes it alongside the stamp.

| Stamp | Flag | Meaning | Action |
|---|---|---|---|
| present | present | legacy upgrade by code update — no activation hook fires | **backfill** |
| absent | absent | genuinely fresh, or mid-activation | **backfill** |
| absent | present | stamp loss on an install that had one | **refuse** |

Verified back to the earliest tags: `v1.2.0` and `v1.2.1` both stamp on every request via `Plugin::init()`, so there is no cohort holding the flag that never received a stamp. "State without stamp is stamp loss" is established, not assumed.

## Both halves are existence tests

`get_option()` returns its default **only when the row is absent**. An earlier cut compared the stamp against the `'0.0.0'` sentinel — so a stamp *emptied* rather than deleted reads `''`, failed the comparison, skipped the guard, and `version_compare( '', WP_SUDO_VERSION, '>=' )` was false so every routine replayed into the re-grant anyway.

Emptying a row is exactly what a migration tool does. Caught in review; now has its own regression test, **verified to fail against unguarded HEAD** in a properly isolated worktree.

## New Site Health critical

**"No user holds the Sudo governance capability."**

Refusing to re-grant is correct; refusing *silently* would leave an operator locked out of Settings → Sudo with nothing to read.

Computed **live** rather than recorded at the moment of refusal. The design review suggested a stored flag; a live check is better here:

- The guard only changes the outcome when holders are zero — which is exactly what this reports — so no information is lost.
- It also catches the commonest real path, the **last holder deleted as a user**, which no upgrade touches.
- It avoids a fourth persistent option, which would cost `PersistentOptionScannerTest`, a tracked metrics row, and both `uninstall.php` cleanup blocks.

Site Health is gated on `view_site_health_checks`, not `manage_wp_sudo` — so a locked-out administrator can actually read the message. Multisite is exempt: `wp_sudo_can()` short-circuits for super admins.

## Accepted cost, chosen rather than inherited

A site stamped **below 3.3.0** whose stamp is *emptied* loses a backfill it genuinely never received. That is indistinguishable from the replay this refuses, and the Site Health status makes the resulting state legible and recoverable.

**Not covered:** whole-namespace loss. An install that loses every `wp_sudo_*` option is indistinguishable from a fresh one by any option-based signal.

## The design review earned its place before any code

Recorded on [#404](https://github.com/dknauss/Sudo/issues/404#issuecomment-5115268716). It corrected two things in my plan:

- **The discriminator.** I intended `wp_sudo_settings` — written on first settings *save*, so it would have **missed replays** on exactly the installs least likely to have opened Settings.
- **The visibility mechanism.** A bare `do_action()` fires into a void: `Event_Recorder::init()` runs *after* `maybe_upgrade()` in `Plugin::init()`, and on the activation path `Plugin::init()` never runs at all.

It also surfaced [#524](https://github.com/dknauss/Sudo/issues/524) — the fresh-activation ordering bug, fixed separately in #531.

## Composition with #531

#531 reorders `activate()` but leaves `update_option( 'wp_sudo_activated', true )` as its final statement, so the fresh-activation and CLI-without-`--user` cases still reach the backfill. The two changes compose; verified by the reviewer against #531's diff.

## Deferred deliberately

Docs beyond CHANGELOG and `release-status.md`. Several items rewrite the same `developer-reference.md` section as #531; writing text now that conflicts on rebase is worse than sweeping once after it merges. Two known nits are on that list: a stale "(defaults to 0.0.0)" parenthetical in a test comment, and a CHANGELOG "since v1.2.1" that *understates* (v1.2.0 has it too).

## Verification

`composer test:unit` 1314 tests / 3915 assertions · `lint` 24/24 · PHPStan level 6 + Psalm clean · `verify:metrics` in sync · `verify:i18n` in sync · `verify:sources` 67 citations present.

The reviewer independently confirmed no legitimate stored stamp can read as `false`, `''` or `'0.0.0'` by reading the `Version:` header out of **every tag** from 1.2.0 forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,42 @@
   *replayed* migration after the version stamp is lost. Nothing asserted this
   before: the existing test checked only that the activating admin received the
   capabilities, which was true either way.
+- **A replayed migration no longer re-grants governance to every administrator
+  (#404).** `Upgrader::upgrade_3_3_0()` bootstraps the four `GOVERNANCE_CAPS` when it
+  finds no holder of `manage_wp_sudo`. It could not tell *"this install predates
+  governance caps"* from *"this install's version stamp went missing"* — both look
+  like zero holders — so a stamp lost to a restore, a staging-to-production copy, a
+  migration tool, or an operator clearing the row replayed every routine from
+  `0.0.0` and handed governance back to every administrator, with no audit trail.
+
+  The routine now refuses that case. `wp_sudo_activated` separates the two:
+  `Plugin::activate()` has written it since v1.2.1 and writes it *after* the
+  upgrader, and `uninstall.php` deletes it alongside the stamp. So a missing stamp
+  on an install that has the flag is stamp **loss**, never a legacy install and
+  never a fresh one. A legacy upgrade by code update — the cohort the routine exists
+  for, where no activation hook fires — keeps its backfill, as does a genuinely
+  fresh activation and a CLI activation without `--user`.
+
+  Both halves of the test are **existence checks**, not sentinel comparisons.
+  `get_option()` returns its default only when the row is absent, so a stamp
+  *emptied* rather than deleted reads `''` — an earlier cut compared against
+  `'0.0.0'` and that request walked past the guard into the re-grant. Emptying a row
+  is exactly what a migration tool does.
+
+  **What it does not cover:** whole-namespace loss. An install that loses every
+  `wp_sudo_*` option is indistinguishable from a fresh one by any option-based
+  signal, and will be re-granted.
+
+  **New Site Health status — "No user holds the Sudo governance capability"
+  (critical).** Refusing to re-grant is correct; refusing *silently* would leave an
+  operator locked out of Settings → Sudo with nothing to read. The status is
+  computed live rather than recorded at the moment of refusal, so it also catches
+  the commonest real path — the last holder being deleted as a user — which no
+  upgrade touches. It names `WP_SUDO_RECOVERY_MODE` as the way back, and says
+  plainly that administrators do not inherit the capability. Site Health is gated on
+  `view_site_health_checks` rather than `manage_wp_sudo`, so a locked-out
+  administrator can actually read it. Multisite is exempt: `wp_sudo_can()`
+  short-circuits for super admins.
 
 ## 4.9.2 - 2026-07-29
 
--- a/docs/current-metrics.md
+++ b/docs/current-metrics.md
@@ -2,15 +2,15 @@
 
 This file is the single source of truth for current repository counts.
 
-Last verified: 2026-07-28
+Last verified: 2026-07-29
 Verification environment: local workspace, PHP 8.x
 
 ## Test Metrics
 
 | Metric | Value | Verification |
 |---|---:|---|
-| Unit tests | 1,308 tests | `composer test:unit` |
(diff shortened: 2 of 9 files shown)


dknauss:
Closing as not planned because the WP Sudo program is concluded. This production-maintenance fix is preserved in the PR history, but no supported or installable successor release will consume it. PR #545 is the canonical conclusion record.

